PR #1, opened 2023-11-29
feat: Initial PgQueue implementation

tomasfarias:
I've pushed two implementations of `PgQueue` in this PR:
* `dequeue` that returns a `PgJob`:
  * Just a normal dequeue, nothing is being held or locked.
  * When moving the job to a new state (`RetryableJob`, `CompletedJob` or `FailedJob`), we also update the db.
* `dequeue_tx` that returns a `PgTransactionJob`:
  * This implementation will begin a transaction and hold the transaction open.
  * When moving the job to a new state (`RetryableJob`, `CompletedJob` or `FailedJob`), we also update the db, and commit the transaction.

`dequeue_tx` has the potential nice benefit of removing all or most of the need for clean-up services: if a worker is terminated, the transaction will be rolledback when postgres timesout, and a new worker can pick-up the job that has been rolledback to `'available'` status. The tradeoffs are several:
1. Each held transaction will prevent postgres VACCUM from cleaning up the row, so we will accumulate more size on the table.
2. More resources (CPU) on Postgres to hold all of those transactions.
3. Lock contention, *but* not with other workers as they run with `SKIP LOCKED`.
4. If we forget to set `idle_in_transaction_session_timeout` (or another timeout setting) in our postgres db, postgres will not timeout (not high risk, we can set this once and forget).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -30,7 +30,7 @@ jobs:
         key: ${{ runner.os }}-cargo-release-${{ hashFiles('**/Cargo.lock') }}
 
     - name: Run cargo build
-      run: cargo build --all --locked --release 
+      run: cargo build --all --locked --release
 
   test:
     runs-on: buildjet-4vcpu-ubuntu-2204
@@ -42,6 +42,24 @@ jobs:
       uses: dtolnay/rust-toolchain@master
       with:
           toolchain: stable
+
+    - name: Install rust
+      uses: dtolnay/rust-toolchain@master
+      with:
+          toolchain: stable
+
+    - name: Stop/Start stack with Docker Compose
+      shell: bash
+      run: |
+          docker compose -f docker-compose.dev.yml down
+          docker compose -f docker-compose.dev.yml up -d
+
+    - name: Run migrations
+      shell: bash
+      run: |
+          cargo install sqlx-cli --no-default-features --features native-tls,postgres
+          DATABASE_URL=postgres://posthog:posthog@localhost:15432/test_database sqlx database create
+          DATABASE_URL=postgres://posthog:posthog@localhost:15432/test_database sqlx migrate run
 
     - uses: actions/cache@v3
       with:
@@ -76,7 +94,7 @@ jobs:
 
     - name: Run clippy
       run: cargo clippy -- -D warnings
-        
+
   format:
     runs-on: buildjet-4vcpu-ubuntu-2204
 
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -43,16 +43,11 @@ jobs:
       with:
           toolchain: stable
 
-    - name: Install rust
-      uses: dtolnay/rust-toolchain@master
-      with:
-          toolchain: stable
-
     - name: Stop/Start stack with Docker Compose
       shell: bash
       run: |
-          docker compose -f docker-compose.dev.yml down
-          docker compose -f docker-compose.dev.yml up -d
+          docker compose -f docker-compose.yml down
+          docker compose -f docker-compose.yml up -d --wait
 
     - name: Run migrations
       shell: bash
--- a/README.md
+++ b/README.md
@@ -1,6 +1,12 @@
 # rusty-hook
 A reliable and performant webhook system for PostHog
 
+## Requirements
+
+1. [Rust](https://www.rust-lang.org/tools/install).
+2. [sqlx-cli](https://crates.io/crates/sqlx-cli): To setup database and run migrations.
+3. [Docker](https://docs.docker.com/engine/install/) or [podman](https://podman.io/docs/installation) (and [podman-compose](https://github.com/containers/podman-compose#installation)): To setup testing services.
+
 ## Testing
 
 1. Start a PostgreSQL instance:
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ A reliable and performant webhook system for PostHog
 
 1. Start a PostgreSQL instance:
 ```bash
-docker compose -f docker-compose.yml up -d
+docker compose -f docker-compose.yml up -d --wait
 ```
 
 2. Prepare test database:
--- a/README.md
+++ b/README.md
@@ -1,2 +1,21 @@
 # rusty-hook
 A reliable and performant webhook system for PostHog
+
+## Testing
+
+1. Start a PostgreSQL instance:
+```bash
+docker compose -f docker-compose.yml up -d
+```
+
+2. Prepare test database:
+```bash
+export DATABASE_URL=postgres://posthog:posthog@localhost:15432/test_database
+sqlx database create
+sqlx migrate run
+```
+
+3. Test:
+```bash
+cargo test
+```
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,15 @@
+services:
+    db:
+        image: docker.io/library/postgres:16-alpine
+        restart: on-failure
+        environment:
+            POSTGRES_USER: posthog
+            POSTGRES_DB: posthog
+            POSTGRES_PASSWORD: posthog
+        healthcheck:
+            test: ['CMD-SHELL', 'pg_isready -U posthog']
+            interval: 5s
+            timeout: 5s
+        ports:
+            - '15432:5432'
+        command: postgres -c max_connections=1000 -c idle_in_transaction_session_timeout=300000
--- a/hook-common/src/pgqueue.rs
+++ b/hook-common/src/pgqueue.rs
@@ -574,23 +574,28 @@ RETURNING
             .fetch_one(&mut *connection)
             .await;
 
-        let job: Job<J> = match query_result {
-            Ok(j) => j,
-            Err(sqlx::Error::RowNotFound) => return Ok(None),
+        match query_result {
+            Ok(job) => Ok(Some(PgJob {
+                job,
+                table: self.table.to_owned(),
+                connection,
+                retry_policy: self.retry_policy,
+            })),
+
+            // Although connection would be closed once it goes out of scope, sqlx recommends explicitly calling close().
+            // See: https://docs.rs/sqlx/latest/sqlx/postgres/any/trait.AnyConnectionBackend.html#tymethod.close.
+            Err(sqlx::Error::RowNotFound) => {
+                let _ = connection.close().await;
+                Ok(None)
+            }
             Err(e) => {
-                return Err(PgQueueError::QueryError {
+                let _ = connection.close().await;
+                Err(PgQueueError::QueryError {
                     command: "UPDATE".to_owned(),
                     error: e,
                 })
             }
-        };
-
-        Ok(Some(PgJob {
-            job,
-            table: self.table.to_owned(),
-            connection,
-            retry_policy: self.retry_policy,
-        }))
+        }
     }
 
     /// Dequeue a Job from this PgQueue to work on it.
@@ -646,23 +651,21 @@ RETURNING
             .fetch_one(&mut *tx)
             .await;
 
-        let job: Job<J> = match query_result {
-            Ok(j) => j,
-            Err(sqlx::Error::RowNotFound) => return Ok(None),
-            Err(e) => {
-                return Err(PgQueueError::QueryError {
-                    command: "UPDATE".to_owned(),
-                    error: e,
-                })
-            }
-        };
+        match query_result {
+            Ok(job) => Ok(Some(PgTransactionJob {
+                job,
+                table: self.table.to_owned(),
+                transaction: tx,
+                retry_policy: self.retry_policy,
+            })),
 
-        Ok(Some(PgTransactionJob {
-            job,
-            table: self.table.to_owned(),
-            transaction: tx,
-            retry_policy: self.retry_policy,
-        }))
+            // Transaction is rolledback on drop.
+            Err(sqlx::Error::RowNotFound) => Ok(None),
+            Err(e) => Err(PgQueueError::QueryError {
+                command: "UPDATE".to_owned(),
+                error: e,
+            }),
+        }
     }
 
     /// Enqueue a Job into this PgQueue.
--- a/hook-common/src/pgqueue.rs
+++ b/hook-common/src/pgqueue.rs
@@ -74,7 +74,7 @@ pub struct Job<J> {
     /// The current status of the job.
     pub status: JobStatus,
     /// Arbitrary job parameters stored as JSON.
-    pub parameters: sqlx::types::Json<J>,
+    pub parameters: JobParameters<J>,
 }
 
 /// A NewJob to be enqueued into a PgQueue.
@@ -83,7 +83,7 @@ pub struct NewJob<J> {
     pub finished_at: Option<DateTime<Utc>>,
     pub started_at: Option<DateTime<Utc>>,
     pub status: JobStatus,
-    pub parameters: sqlx::types::Json<J>,
+    pub parameters: JobParameters<J>,
 }
 
 impl<J> NewJob<J> {
--- a/hook-common/src/pgqueue.rs
+++ b/hook-common/src/pgqueue.rs
@@ -85,46 +85,40 @@ pub struct Job<J> {
 }
 
 impl<J> Job<J> {
+    /// Consume Job to retry it.
+    /// This returns a RetryableJob that can be enqueued by PgQueue.
+    ///
+    /// # Arguments
+    ///
+    /// * `error`: Any JSON-serializable value to be stored as an error.
     pub fn retry<E: Serialize>(self, error: E) -> Result<RetryableJob<E>, PgQueueError> {
-        if self.attempt == self.max_attempts {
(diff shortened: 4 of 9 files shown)


migrations/20231129172339_job_queue_table.sql
```diff
@@ -0,0 +1,16 @@
...
+);
+
+CREATE TABLE job_queue(
+    id BIGSERIAL PRIMARY KEY,
+    attempt INT NOT NULL DEFAULT 0,
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    finished_at TIMESTAMPTZ DEFAULT NULL,
+    started_at TIMESTAMPTZ DEFAULT NULL,
+    status job_status NOT NULL DEFAULT 'available'::job_status,
+    parameters JSONB
```
davemurphysf: Thinking ahead to tracking/managing correlated failures (i.e. if Salesforce is down, we stop/pause all hooks to Salesforce), do we want to break out URL or host into its own column from the JSON object?

migrations/20231129172339_job_queue_table.sql
```diff
@@ -0,0 +1,16 @@
+CREATE TYPE job_status AS ENUM(
+    'available',
+    'completed',
+    'failed',
+    'running'
+);
+
+CREATE TABLE job_queue(
```
davemurphysf, line 8: How are we tracking next attempt timing (i.e. exponential backoff)?

.github/workflows/rust.yml
```diff
@@ -43,6 +43,24 @@ jobs:
       with:
           toolchain: stable
 
+    - name: Install rust
+      uses: dtolnay/rust-toolchain@master
+      with:
+          toolchain: stable
```
bretthoerner: Seems like a duplicate with the `Install rust` right above.

.github/workflows/rust.yml
```diff
@@ -43,6 +43,24 @@ jobs:
...
+
+    - name: Stop/Start stack with Docker Compose
+      shell: bash
+      run: |
+          docker compose -f docker-compose.dev.yml down
+          docker compose -f docker-compose.dev.yml up -d
```
bretthoerner: I'm not a `docker compose` expert but we might want `--wait` so everything is up before continuing.

README.md
```diff
@@ -1,2 +1,21 @@
 # rusty-hook
 A reliable and performant webhook system for PostHog
+
+## Testing
+
+1. Start a PostgreSQL instance:
+```bash
+docker compose -f docker-compose.yml up -d
```
bretthoerner: Assuming the `--wait` is warranted in the CI steps above,

```suggestion
docker compose -f docker-compose.yml up -d --wait
```

.github/workflows/rust.yml
bretthoerner:
```suggestion
          docker compose -f docker-compose.dev.yml up -d --wait
```

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+services:
+    db:
+        image: docker.io/library/postgres:16-alpine
```
bretthoerner, line 3: More of a question than a review comment, but since I don't think we plan to do transaction enqueueing, I assume we'll run our own RDS instance for this service? In that case we can actually run Postgres 16.

hook-common/src/pgqueue.rs
```diff
@@ -0,0 +1,215 @@
...
+    SELECT
+        id
+    FROM
+        "{0}"
+    WHERE
+        status = 'available'
+    ORDER BY
+        id
+    LIMIT 1
+    FOR UPDATE SKIP LOCKED
```
bretthoerner, line 560: For future readers' sake can you link to https://www.2ndquadrant.com/en/blog/what-is-select-skip-locked-for-in-postgresql-9-5/ (or anything else if you have something better)

hook-common/src/pgqueue.rs
```diff
@@ -0,0 +1,215 @@
...
+pub type JobParameters<J> = sqlx::types::Json<J>;
+
+/// A Job to be executed by a worker dequeueing a PgQueue.
+#[derive(sqlx::FromRow)]
+pub struct Job<J> {
+    pub id: i64,
+    pub attempt: i32,
+    pub finished_at: Option<DateTime<Utc>>,
+    pub created_at: DateTime<Utc>,
+    pub started_at: Option<DateTime<Utc>>,
```
bretthoerner: We can always add it later, so feel free to ignore me, but I think the [attempted_by](https://github.com/riverqueue/river/blob/a8f0ce74e93ee3c73f91b2781cbe0033af2a8554/internal/dbsqlc/river_job.sql#L16) field in River would be super useful for debugging (i.e. it could be the pod name, or something)

bretthoerner (approved):

migrations/20231129172339_job_queue_table.sql
bretthoerner: Yeah, I have a related question, since I wasn't around when we discussed the plans for this.

Do we have any plans for dealing with pathologically slow backends? For example say Slack accepts connections but they all time out, and a lot of our webhooks are configured to send there. It seems like every worker will stall up attempting every Slack hook send. Sure, each individual job may exponentially backoff, but in aggregate the webhook system will act a lot like Kafka with head-of-line blocking in this situation.

I guess it's not something we _require_ for v1, but I'm curious if it was discussed at all.

tomasfarias: It was discussed. The design of the system includes a global management service that will disable bad endpoints by querying PG. This service will disable _all_ webhooks of the same endpoint. The details of the system were not discussed though, but I imagine that we have a service that queries PG every X seconds and if it reads failures for a given endpoint are above a given threshold then we disable it.

However the system is implemented, seems like we have agreed it will be separate from the queue system. But we may need to extend the queue schema to support this, perhaps with an endpoint/destination/plugin_config_id column and a 'disabled' status?

.github/workflows/rust.yml
tomasfarias: We have a separate step in the PostHog pipeline, wonder if that's better specifically for Postgres as even if the db is "ready" it may not be fully initialized.

tomasfarias: But anyways, good observation!

docker-compose.yml
tomasfarias: I opted for the latest version arbitrarily. Regardless of the service providing us with Postgres, the docker-compose was more intended for development, and I planned on  adding an echo server service for sending requests to (whenever we get to that part). 

I don't have an opinion regarding RDS or other Postgres hosting providers. As long as they are quick to spin up (and offer a relatively up-to-date Postgres), I think we don't need much else. Happy to discuss this further.

hook-common/src/pgqueue.rs
tomasfarias: Yeah, works for me, could even be thread id in a single pod multithreaded context.

.github/workflows/rust.yml
tomasfarias: Went with `--wait`!

.github/workflows/rust.yml
tomasfarias: Yup, copy pasting late in the day...

migrations/20231129172339_job_queue_table.sql
tomasfarias: We are not, at least not yet. I'm thinking of introducing a `retryable_at` timestamp field:
* When the job is first inserted, this value is `NULL`, and it's ignored by dequeue.
* When a job fails and has to be retried this can be updated to a time in the future.
  * The time can be decided by a retry policy (i.e. exponential backoff).
* Any dequeue operations will ignore jobs with retryable_at in the future.
* The downside is that we are not guaranteeing a retry will happen in, say, 5 seconds, but that a retry _could_ happen in 5 seconds. For example, it could not happen if no workers are listening on the queue, or it could happen after 5 seconds if things are slow. Still, I think this is reasonable.

I think, we will also need to hold a lock beyond the call to `dequeue` as the worker won't know if it will fail as it's dequeueing the job, it only does so once it's done. So, by opening a transaction and holding a lock while the worker works, all concurrent dequeues will skip this row. Not exactly sure how this implementation will look like, working through some ideas...

Any thoughts?

ellie (approved):

migrations/20231129172339_job_queue_table.sql
tomasfarias: For example: river queue provides a `kind` field that uniquely identifies a type of job. In our context, this could be `endpoint` and uniquely identify a webhook endpoint. Continuing with river queue as an example, they also have a `"discarded"` state to identify jobs that have failed for the last time.

bretthoerner: Cool, that makes sense. And yeah, it does seem like we'll want an easier column to query than the raw JSONB. Could be added in later iterations though, up to you.

docker-compose.yml
bretthoerner: Makes sense, I only brought it up because I feel production and dev Postgres versions should match when we're getting into nitty gritty Postgres feature specifics like we will in this repo. 🤷

tomasfarias: Oh yeah, of course. I feel like we are not using anything particularly new from Postgres, so we can shift the docker-compose version around depending on our hosting solution.

tomasfarias: I'll bring up the conversation, thank you for raising this point @bretthoerner

migrations/20231129172339_job_queue_table.sql
bretthoerner: I'm curious what River does. It seems like it differentiates `created_at` from `scheduled_at` which seems a bit like your proposal? I haven't read all of their code, but I assume queue workers would only need to query by `scheduled_at` (which starts out equal to `created_at` anyway).

If a Worker task fails, I assume it updates the job's `scheduled_at` based on the backoff logic.

And if a Worker dies, we need some periodic background progress that looks for very stale `attempted_at` and assumes (and/or verifies) the Worker is dead and updates the rows with the new proper `scheduled_at`.

I'm mostly riffing here, but it seems like this avoids any holding-open-of-transactions?

README.md
```diff
@@ -1,2 +1,21 @@
...
+
+1. Start a PostgreSQL instance:
+```bash
+docker compose -f docker-compose.yml up -d --wait
+```
+
+2. Prepare test database:
+```bash
+export DATABASE_URL=postgres://posthog:posthog@localhost:15432/test_database
+sqlx database create
```
tiina303, line 20: maybe some deps instructions beforehand too? for sqlx and cargo

migrations/20231129172339_job_queue_table.sql
```diff
@@ -0,0 +1,18 @@
...
+    id BIGSERIAL PRIMARY KEY,
+    attempt INT NOT NULL DEFAULT 0,
+    attempted_at TIMESTAMPTZ DEFAULT NULL,
+    attempted_by TEXT[] DEFAULT ARRAY[]::TEXT[],
+    created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+    finished_at TIMESTAMPTZ DEFAULT NULL,
+    started_at TIMESTAMPTZ DEFAULT NULL,
```
tiina303: how are we tracking how long we waited for the endpoint - e.g. if someone is close to the endpoint being too slow we could warn them in the UI and also useful info for us in case of lag in processing 🤷‍♀️

migrations/20231129172339_job_queue_table.sql
tomasfarias: I'll open a follow-up PR to discuss retries (and with a first implementation), just to keep things separate.

README.md
tomasfarias: Yeah, that would go in a first section on how to set-up the project

migrations/20231129172339_job_queue_table.sql
tomasfarias: It's implicitly tracked by `started_at - finished_at`.

tiina303: ack, so those are both for the actual fetch call, not when we take something from the queue to be processed

migrations/20231129172339_job_queue_table.sql
davemurphysf: > I think, we will also need to hold a lock beyond the call to dequeue as the worker won't know if it will fail as it's dequeueing the job, it only does so once it's done. So, by opening a transaction and holding a lock while the worker works, all concurrent dequeues will skip this row. 

This was my understanding as well

tomasfarias: > I'm curious what River does. It seems like it differentiates created_at from scheduled_at which seems a bit like your proposal? 

It's similar: River supports scheduling jobs in the future, regardless if they are being retried. That's why they have an additional "scheduled" state that a job can start as besides "available".

We do not support scheduling: Webhooks are enqueued to be run "now" (or relatively close to "now"). However, if you need to retry a job, it works as if you were scheduling it for the future. Hence calling it `retryable_at` instead of `scheduled_at`, but `scheduled_at` would be fine too.

README.md
tomasfarias: I've added a list of requirements just above this section with links to installation guides for each of them.

migrations/20231129172339_job_queue_table.sql
tomasfarias: I went with `scheduled_at` for our retry implementation, just in case we ever want to support scheduling webhooks, and because the name still works with normal retrying.

migrations/20231129172339_job_queue_table.sql
tomasfarias: I didn't open a follow-up PR and instead implemented it here...

migrations/20231129172339_job_queue_table.sql
tomasfarias: `started_at` (now renamed to `attempted_at` in my latest commit to better differentiate it from `created_at`) and `finished_at` capture whatever time it took the worker to do whatever work it set out to do. If the work is making a fetch call, then that's what we are recording, but the queue is work-agnostic.

hook-common/src/pgqueue.rs
```diff
@@ -0,0 +1,926 @@
...
+                id: self.id,
+                attempt: self.attempt,
+                error: sqlx::types::Json(error),
+                queue: self.queue,
+            })
+        }
+    }
+
+    /// Consume Job to complete it.
+    /// This returns a CompletedJob that can be enqueued by PgQueue.
```
bretthoerner:
```suggestion
    /// This returns a CompletedJob that can be marked as completed by PgQueue.
```

hook-common/src/pgqueue.rs
```diff
@@ -0,0 +1,926 @@
...
+    pub fn complete(self) -> CompletedJob {
+        CompletedJob {
+            id: self.id,
+            queue: self.queue,
+        }
+    }
+
+    /// Consume Job to fail it.
+    /// This returns a FailedJob that can be enqueued by PgQueue.
```
bretthoerner:
```suggestion
    /// This returns a FailedJob that can be marked as failed by PgQueue.
```

hook-common/src/pgqueue.rs
```diff
@@ -0,0 +1,926 @@
...
+            parameters: sqlx::types::Json(parameters),
+            target: target.to_owned(),
+        }
+    }
+}
+
+#[derive(Copy, Clone)]
+/// The retry policy that PgQueue will use to determine how to set scheduled_at when enqueuing a retry.
+pub struct RetryPolicy {
+    /// Coeficient to multiply initial_interval with for every past attempt.
```
bretthoerner:
```suggestion
    /// Coefficient to multiply initial_interval with for every past attempt.
```

Totally unimportant. :)

hook-common/src/pgqueue.rs
```diff
@@ -0,0 +1,926 @@
...
+    /// The identifier of the worker listening on this queue.
+    worker: String,
+}
+
+pub type PgQueueResult<T> = std::result::Result<T, PgQueueError>;
+
+impl PgQueue {
+    /// Initialize a new PgQueue backed by table in PostgreSQL.
+    pub async fn new(
+        name: &str,
```
bretthoerner: Minor, but this may be better as `queue` or `queue_name` or something, if I understand what it's used for correctly.

bretthoerner (approved):
Just re-read, looking good. 👍

hook-common/src/pgqueue.rs
tomasfarias: It's passing a name for the queue when initializing a new one. Renaming to `queue_name` works for me too :+1:

tomasfarias: Don't know if we'll need multiple queues on the table, but it was easy to implement (one extra predicate in the `WHERE`) and helped with tests to have different queues (one for each test).

tomasfarias: I've also renamed `table` and `worker` to `table_name`, `worker_name` to keep in line. Also added documentation.

PR #2, opened 2023-12-04
Setup basic API

ellie:
Needs routes to hook into pgqueue obv, but

1. Pull out common dependencies, either currently common or likely future common
2. Add http metrics
3. Start up axum server
4. Add config

### DIFF
--- a/hook-producer/src/metrics.rs
+++ b/hook-producer/src/metrics.rs
@@ -1,0 +1,53 @@
+use std::time::Instant;
+
+use axum::{
+    body::Body, extract::MatchedPath, http::Request, middleware::Next, response::IntoResponse,
+};
+use metrics_exporter_prometheus::{Matcher, PrometheusBuilder, PrometheusHandle};
+
+pub fn setup_metrics_recorder() -> PrometheusHandle {
+    const EXPONENTIAL_SECONDS: &[f64] = &[
+        0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0,
+    ];
+
+    PrometheusBuilder::new()
+        .set_buckets_for_metric(
+            Matcher::Full("http_requests_duration_seconds".to_string()),
+            EXPONENTIAL_SECONDS,
+        )
+        .unwrap()
+        .install_recorder()
+        .unwrap()
+}
+
+/// Middleware to record some common HTTP metrics
+/// Someday tower-http might provide a metrics middleware: https://github.com/tower-rs/tower-http/issues/57
+pub async fn track_metrics(req: Request<Body>, next: Next) -> impl IntoResponse {
+    let start = Instant::now();
+
+    let path = if let Some(matched_path) = req.extensions().get::<MatchedPath>() {
+        matched_path.as_str().to_owned()
+    } else {
+        req.uri().path().to_owned()
+    };
+
+    let method = req.method().clone();
+
+    // Run the rest of the request handling first, so we can measure it and get response
+    // codes.
+    let response = next.run(req).await;
+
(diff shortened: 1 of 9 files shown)


hook-producer/src/metrics.rs
```diff
@@ -0,0 +1,53 @@
...
+            Matcher::Full("http_requests_duration_seconds".to_string()),
+            EXPONENTIAL_SECONDS,
+        )
+        .unwrap()
+        .install_recorder()
+        .unwrap()
+}
+
+/// Middleware to record some common HTTP metrics
+/// Someday tower-http might provide a metrics middleware: https://github.com/tower-rs/tower-http/issues/57
```
bretthoerner, line 24: Nice, thanks for the link, subscribed.

hook-producer/src/metrics.rs
```diff
@@ -0,0 +1,53 @@
...
+        matched_path.as_str().to_owned()
+    } else {
+        req.uri().path().to_owned()
+    };
+
+    let method = req.method().clone();
+
+    // Run the rest of the request handling first, so we can measure it and get response
+    // codes.
+    let response = next.run(req).await;
```
bretthoerner, line 38: Huh, the way they do middleware is cool.

bretthoerner (approved):

bretthoerner:
Hm, I didn't expect the `Unresolved conversations` checks. That seems kind of annoying.

hook-producer/src/metrics.rs
```diff
@@ -0,0 +1,53 @@
...
+
+    let latency = start.elapsed().as_secs_f64();
+    let status = response.status().as_u16().to_string();
```
tomasfarias, line 41: I really wonder why `as_str` goes through the trouble of [figuring an offset](https://docs.rs/http/latest/src/http/status.rs.html#146) to get the string representation of the status code from a [hardcoded string of codes](https://docs.rs/http/latest/src/http/status.rs.html#543).

tomasfarias: Maybe it's so it doesn't need to allocate an additional `String`?

tomasfarias: Anyways, this is perfectly fine.

hook-producer/src/metrics.rs
```diff
@@ -0,0 +1,53 @@
+use std::time::Instant;
+
+use axum::{
+    body::Body, extract::MatchedPath, http::Request, middleware::Next, response::IntoResponse,
+};
+use metrics_exporter_prometheus::{Matcher, PrometheusBuilder, PrometheusHandle};
+
+pub fn setup_metrics_recorder() -> PrometheusHandle {
+    const EXPONENTIAL_SECONDS: &[f64] = &[
+        0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0,
```
tomasfarias, line 10: question: Where does this "x2/x2.5" series come from?

hook-producer/src/metrics.rs
ellie: no problem

ellie:
> Hm, I didn't expect the Unresolved conversations checks. That seems kind of annoying.

Oof yeah, I'll disable it

hook-producer/src/metrics.rs
ellie: It's a series I've copy-pasted from my [snippet](https://ellie.wtf/notes/exporting-prometheus-metrics-with-axum) for a while now, I think originating from the axum example docs. Generally seems to work quite well, no issues if it needs changing.

tomasfarias: Cool! Thank you! I was just curious where it came from. I think it works fine!

PR #4, opened 2023-12-07
Add basic produce endpoint

bretthoerner:


hook-common/src/pgqueue.rs
```diff
@@ -560,9 +688,8 @@ impl PgQueue {
         let name = queue_name.to_owned();
         let table = table_name.to_owned();
         let pool = PgPoolOptions::new()
-            .connect(url)
-            .await
-            .map_err(|error| PgQueueError::ConnectionError { error })?;
+            .connect_lazy(url)
```
bretthoerner, line 568: As far as switching to `connect_lazy`: my gut feeing is that PG can be down any time, and we shouldn't crash-loop on startup if it's unavailable. It seems better to always present the same API to users, if PG is unavailable maybe we 500, but eventually things will work again.

Not to mention crash-looping in k8s will eventually give up. Seems better to leave the process up and let it keep trying?

hook-common/src/pgqueue.rs
```diff
@@ -528,6 +655,7 @@ impl Default for RetryPolicy {
 }
 
 /// A queue implemented on top of a PostgreSQL table.
+#[derive(Clone)]
```
bretthoerner, line 535: `with_state(...)` (see `Router`) appears to be how Axum handles passing non-static state into endpoints. They need to be `Copy`, and cloning `PgPool` is cheap. Cloning the strings seems like a waste but is definitely not going to be on our radar anytime soon... but what we put in here is something we need to keep an eye on.

Unless I'm missing a better way to make this available in my producer endpoints?

hook-producer/src/handlers/mod.rs
```diff
@@ -1,15 +1,164 @@
...
+pub fn app(pg_pool: PgQueue, metrics: Option<PrometheusHandle>) -> Router {
     Router::new()
         .route("/", routing::get(index::get))
         .route(
             "/metrics",
-            routing::get(move || std::future::ready(recorder_handle.render())),
+            routing::get(move || match metrics {
+                Some(ref recorder_handle) => std::future::ready(recorder_handle.render()),
+                None => std::future::ready("no metrics recorder installed".to_owned()),
+            }),
```
bretthoerner: Metrics can only be installed once, so the pre-existing code fucked up tests (more than 1 test = panic). I made it optional so tests don't have to set it up at all.

hook-producer/src/handlers/mod.rs
```diff
@@ -1,15 +1,164 @@
...
+                    .uri("/webhook")
+                    .header(http::header::CONTENT_TYPE, "application/json")
+                    .body("{}".to_owned())
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+
+        assert_eq!(response.status(), StatusCode::UNPROCESSABLE_ENTITY);
+    }
```
bretthoerner: I can definitely add more test coverage given time, I just got to the basics today.

hook-producer/src/main.rs
```diff
@@ -21,10 +21,24 @@ async fn listen(app: Router, bind: String) -> Result<()> {
 async fn main() {
     tracing_subscriber::fmt::init();
 
-    let app = handlers::router();
-
     let config = Config::init_from_env().expect("failed to load configuration from env");
 
+    let pg_queue = PgQueue::new(
+        // TODO: Coupling the queue name to the PgQueue object doesn't seem ideal from the producer
+        // side, but we don't need more than one queue for now.
```
tomasfarias, line 28: We actually need two for the v1 spec, so we may address this soon.

hook-producer/src/main.rs
```diff
@@ -21,10 +21,24 @@ async fn listen(app: Router, bind: String) -> Result<()> {
...
+        &config.queue_name,
+        &config.table_name,
+        &config.database_url,
+        // TODO: It seems unnecessary that the producer side needs to know about the retry policy.
+        RetryPolicy::default(),
```
tomasfarias, line 33: Yeah, we can move this out of the queue and only on the consumer.

hook-producer/src/handlers/webhook.rs
```diff
@@ -0,0 +1,63 @@
...
+
+pub async fn post(
+    State(pg_queue): State<PgQueue>,
+    Json(payload): Json<WebhookJobParameters>,
+) -> Result<Json<WebhookPostResponse>, (StatusCode, Json<WebhookPostResponse>)> {
+    debug!("received payload: {:?}", payload);
+
+    // TODO: This needs to be configurable, but shouldn't `max_attempts` be embedded in the PgQueue
+    // object? Or is there a reason we have to supply it manually for each `enqueue`?
+    let max_attempts = 3;
```
tomasfarias: I'd argue `max_attempts` is a property of the job, so whoever is enqueueing a job needs to define it. Maybe plugin-server should give us this?

tomasfarias: We can have a default value for all though. I don't see us requiring different number of `max_attempts` for different jobs.

hook-producer/src/handlers/mod.rs
tomasfarias: You can also use conditional compilation (with cfg!(test) or #[cfg(test)]) to avoid introducing runtime optional handling (even when it should always be defined in release).

hook-producer/src/handlers/webhook.rs
```diff
@@ -0,0 +1,63 @@
...
+    })?;
+    let url_hostname = match url.host_str() {
+        Some(hostname) => hostname,
+        None => {
+            error!("couldn't extract hostname from url: {}", payload.url);
+            return Err((
+                StatusCode::INTERNAL_SERVER_ERROR,
+                Json(WebhookPostResponse {
+                    error: Some("couldn't extract hostname from url".to_owned()),
+                }),
```
tomasfarias: Shouldn't this also be a bad request? Making it a 5XX kind of implies that we _want_ to support URLs without a host, but something went wrong on our end. However, I don't see us supporting any of the examples without a host I'm seeing here: https://docs.rs/url/latest/url/struct.Url.html#method.host_str

hook-producer/src/handlers/webhook.rs
tomasfarias: Happy to introduce a bit of builder pattern for this.

hook-common/src/pgqueue.rs
tomasfarias: The pool itself is indeed cheap to clone. We can use a ref count pointer if the rest of this struct ever gets expensive to clone: https://docs.rs/axum/latest/axum/index.html#using-the-state-extractor

hook-common/src/pgqueue.rs
tomasfarias: Yeah, fair enough we can connect later :+1:

hook-common/src/pgqueue.rs
```diff
@@ -67,6 +74,126 @@ impl FromStr for JobStatus {
...
+            HttpMethod::PUT => http::Method::PUT,
+        }
+    }
+}
+
+/// `JobParameters` required for the `WebhookConsumer` to execute a webhook.
+/// These parameters should match the exported Webhook interface that PostHog plugins.
+/// implement. See: https://github.com/PostHog/plugin-scaffold/blob/main/src/types.ts#L15.
+#[derive(Deserialize, Serialize, Debug, PartialEq, Clone)]
+pub struct WebhookJobParameters {
```
tomasfarias: I rather keep the queue job-agnostic. Sharing the parameters between consumer and producer makes sense to ensure we are both working with the same structs, but can we move this to a separate module? Maybe `hook-common/src/webhook.rs`?

hook-producer/src/handlers/webhook.rs
tomasfarias: I would consider putting the `Url::parse` and `host_str()` calls in a separate function that returns a hostname or a 400 error.

hook-producer/src/handlers/webhook.rs
```diff
@@ -0,0 +1,63 @@
...
+
+#[derive(Serialize)]
+pub struct WebhookPostResponse {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    error: Option<String>,
+}
+
+pub async fn post(
+    State(pg_queue): State<PgQueue>,
+    Json(payload): Json<WebhookJobParameters>,
```
tomasfarias, line 20: question: How does axum react if the json payload is malformed (i.e. cannot be deserialized).

tomasfarias: Looking for a way to enforce payload size limits, and we could do this by changing the deserialization to fail if size limits are exceeded, but curious what error will the user see from this.

hook-producer/src/handlers/mod.rs
bretthoerner: Thanks, I checked that out but `if cfg!(test)` seems like it would introduce the same `Some` vs `None` code that I need to have here.

I guess it would let me optionally construct the metrics thing inside the `app()` function, but I kind of like to create these dependencies in `main` anyway. Happy to change it if you think I should.

hook-producer/src/handlers/webhook.rs
bretthoerner: axum returns the proper `4XX` code depending on what failed, I'll add some more tests showing that.

If you're worried about the client hurting the server (i.e. sending huuuuge body) then I guess we need to validate that via serde and also verify it short-circuits properly. Although we own the clients so I think that's less of a fear.

If we want to keep it simple we can just do a length check on the `body` `String` and toss out a `400`. I may go ahead and add that here and we can adjust it as desired later?

hook-producer/src/handlers/webhook.rs
bretthoerner: Good call, I'll just add it to the payload with a default, and that'll solve both.

hook-producer/src/main.rs
bretthoerner: Cool, it's not a big deal or anything, I'll leave the TODO for later.

bretthoerner:
Going to add some more 4XX tests but this is ready otherwise.

hook-common/src/webhook.rs
```diff
@@ -0,0 +1,134 @@
+use std::collections;
```
bretthoerner, line 1: This file is all code moves.

hook-producer/src/handlers/webhook.rs
tomasfarias: Yeah, I think that's good enough to get us started :+1: We own the clients, but we don't do any validation on them afaik.

tomasfarias (approved):
lgtm, thanks for the changes!

hook-producer/src/handlers/mod.rs
tomasfarias: We can roll it out like this and if it becomes to annoying revisit it in the future :+1:

PR #5, opened 2023-12-08
refactor: Re-use client in consumer

tomasfarias:
Follow-up to #3.

This PR implements re-using the `reqwest::Client` spawned in consumer instead of creating a new one for every request. This allows us to:
* Overall, be more efficient.
* Set timeout once for all requests.
* Set headers once for all requests.

This PR also applies a limit to the maximum number of concurrent tasks using a Semaphore. Set to 1024 as default. Follows this pattern: https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#limit-the-number-of-incoming-requests-being-handled-at-the-same-time

Other changes:
* Moved `WebhookConsumerError` to its own module.

### DIFF
--- a/hook-consumer/src/config.rs
+++ b/hook-consumer/src/config.rs
@@ -21,7 +21,7 @@ pub struct Config {
     pub request_timeout: EnvMsDuration,
 
     #[envconfig(default = "1024")]
-    pub max_requests: usize,
+    pub max_concurrent_jobs: usize,
 
     #[envconfig(nested = true)]
     pub retry_policy: RetryPolicyConfig,
--- a/hook-consumer/src/config.rs
+++ b/hook-consumer/src/config.rs
@@ -20,6 +20,9 @@ pub struct Config {
     #[envconfig(default = "5000")]
     pub request_timeout: EnvMsDuration,
 
+    #[envconfig(default = "1024")]
+    pub max_requests: usize,
+
     #[envconfig(nested = true)]
     pub retry_policy: RetryPolicyConfig,
 
--- a/hook-consumer/src/consumer.rs
+++ b/hook-consumer/src/consumer.rs
@@ -6,34 +6,10 @@ use std::time;
 use async_std::task;
 use hook_common::pgqueue::{PgJobError, PgQueue, PgQueueError, PgTransactionJob};
 use http::StatusCode;
+use reqwest::header;
 use serde::{de::Visitor, Deserialize, Serialize};
-use thiserror::Error;
 
-/// Enumeration of errors for operations with WebhookConsumer.
-#[derive(Error, Debug)]
-pub enum WebhookConsumerError {
-    #[error("timed out while waiting for jobs to be available")]
-    TimeoutError,
-    #[error("{0} is not a valid HttpMethod")]
-    ParseHttpMethodError(String),
-    #[error("error parsing webhook headers")]
-    ParseHeadersError(http::Error),
-    #[error("error parsing webhook url")]
-    ParseUrlError(url::ParseError),
-    #[error("an error occurred in the underlying queue")]
-    QueueError(#[from] PgQueueError),
-    #[error("an error occurred in the underlying job")]
-    PgJobError(String),
-    #[error("an error occurred when attempting to send a request")]
-    RequestError(#[from] reqwest::Error),
-    #[error("a webhook could not be delivered but it could be retried later: {reason}")]
-    RetryableWebhookError {
-        reason: String,
-        retry_after: Option<time::Duration>,
-    },
-    #[error("a webhook could not be delivered and it cannot be retried further: {0}")]
-    NonRetryableWebhookError(String),
-}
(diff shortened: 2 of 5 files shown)


hook-consumer/src/config.rs
```diff
@@ -20,6 +20,9 @@ pub struct Config {
     #[envconfig(default = "5000")]
     pub request_timeout: EnvMsDuration,
 
+    #[envconfig(default = "1024")]
+    pub max_requests: usize,
```
bretthoerner: I'd call it `max_concurrent_requests` or `max_outstanding_requests` or something just to avoid any ambiguity.

hook-consumer/src/consumer.rs
```diff
@@ -199,12 +192,20 @@ impl<'p> WebhookConsumer<'p> {
 
     /// Run this consumer to continuously process any jobs that become available.
     pub async fn run(&self) -> Result<(), WebhookConsumerError> {
+        let semaphore = Arc::new(sync::Semaphore::new(self.max_concurrent_jobs));
+
         loop {
-            // TODO: The number of jobs processed will be capped by the PG connection limit when running in transactional mode.
```
tomasfarias, line 203: I've also removed this comment as `max_concurrent_jobs` now caps the processing even without PG connection limit.

hook-consumer/src/consumer.rs
```diff
@@ -199,12 +192,20 @@ impl<'p> WebhookConsumer<'p> {
...
-            let request_timeout = self.request_timeout; // Required to avoid capturing self in closure.
-            tokio::spawn(async move { process_webhook_job(webhook_job, request_timeout).await });
+            // reqwest::Client internally wraps with Arc, so this allocation is cheap.
+            let client = self.client.clone();
+            let permit = semaphore.clone().acquire_owned().await.unwrap();
+
+            tokio::spawn(async move {
+                let result = process_webhook_job(client, webhook_job).await;
+                drop(permit);
+                result.expect("webhook processing failed");
```
bretthoerner: Doesn't `expect` on an Error panic? If I'm reading right, this will Error if a PG query fails, but I don't think we want to crash the process in that case do we?

hook-consumer/src/config.rs
tomasfarias: Oh, before I read your comment I changed the name to `max_concurrent_jobs` in this commit: https://github.com/PostHog/rusty-hook/pull/5/commits/31b1e7ae0e6f19fcf9cf40fd0e2a87e779667f19. What do you think about that?

tomasfarias: I originally thought about capping only the HTTP requests to avoid hitting the file descriptor limit, hence the original name.

hook-consumer/src/consumer.rs
tomasfarias: Yeah, it will panic, but inside a task (thread), which should not propagate to the main thread: https://users.rust-lang.org/t/panic-in-tokio-task-does-not-end-the-program-execution/45731/2

tomasfarias: Should be fine to return an error too, we are not doing any handling anyways (and that's something that we should, but thought about focusing on tracking/graceful handling afterwards).

bretthoerner: > Yeah, it will panic, but inside a task (thread), which should not propagate to the main thread: [users.rust-lang.org/t/panic-in-tokio-task-does-not-end-the-program-execution/45731/2](https://users.rust-lang.org/t/panic-in-tokio-task-does-not-end-the-program-execution/45731/2)

Ahhh, TIL, thanks.

Do we benefit somehow from panicking the task? Returning an error just seems easier for future readers (me, lol).

hook-consumer/src/config.rs
bretthoerner: > Oh, before I read your comment I changed the name to `max_concurrent_jobs` in this commit: [31b1e7a](https://github.com/PostHog/rusty-hook/commit/31b1e7ae0e6f19fcf9cf40fd0e2a87e779667f19). What do you think about that?

Sounds good to me.

hook-consumer/src/consumer.rs
tomasfarias: Not really, I was just being lazy and explicitly saying "if we get here we are fucked". Eventually we should think about error handling, so I changed it to return the error.

tomasfarias: Handling gets tricky as if we get to that point it means something is wrong with the database...

bretthoerner: > Handling gets tricky as if we get to that point it means something is wrong with the database...

That's true, but as currently written we'd drop the txn (or try, it might have disconnected which would do the same thing), release the row, and it would (eventually) retry when the DB worked again right? So everything would stall (as expected) but we wouldn't really lose anything, right? Seems like the behavior we want?

bretthoerner (approved):

PR #7, opened 2023-12-12
Add indexes, drop redundant column

bretthoerner:


migrations/20231212184028_indexes.sql
```diff
@@ -0,0 +1,9 @@
+-- Redundant with finished_at
+ALTER TABLE
+    job_queue DROP COLUMN completed_at;
+
+-- Needed for `dequeue` queries
+CREATE INDEX idx_job_queue_status_scheduled_at_queue ON job_queue(status, scheduled_at, queue);
+
+-- Needed for UPDATE-ing incomplete jobs with a specific target (i.e. slow destinations)
+CREATE INDEX idx_job_queue_status_target_queue ON job_queue(status, target, queue);
```
tomasfarias: I think I would rather have `queue` as the left-most column: Consumers (as of right now) can only listen on one queue, so will always filter on `queue = $queue_name`. Given that constraint will always be an equality constraint, we should keep it in the leading columns to ensure PG uses it (whereas `scheduled_at` for example will likely be a inequality)

migrations/20231212184028_indexes.sql
```diff
@@ -0,0 +1,9 @@
+-- Redundant with finished_at
+ALTER TABLE
+    job_queue DROP COLUMN completed_at;
```
tomasfarias: nit(naming): Should we keep `completed_at` and drop `finished_at`? Any thoughts into which name fits better?

tomasfarias: I'm kind of indifferent, but wondering if there's any reason to prefer one over the other.

migrations/20231212184028_indexes.sql
bretthoerner: Interesting, I did some searches o refresh myself when I wrote this, and the advice seemed to be "put the most restrictive filter first", and I figured the table would typically be full of `completed` events, or retry events that are in the future, and so `(status, scheduled_at)` would tend to get to a small number of rows that `queue` would then need to filter.

Of course I didn't do any testing or anything, and I'm not sure it's worth taking that detour right now since we can also just change the index around and see how it goes pretty easily.

Either way, I'm totally fine changing it, just wanted to write down what I was thinking.

migrations/20231212184028_indexes.sql
bretthoerner: As it's _currently_ used, `finished_at` feels slightly more correct. Because we set it even when we put a row back in for a `retry`:

https://github.com/PostHog/rusty-hook/blob/126908fd4f57b5b85bb6b618b4229a17e67d892b/hook-common/src/pgqueue.rs#L285-L316

To me, `completed` really means we're done (`completed` or `failed`) where it seems like we're using `finished_at` as more like a `last_attempt_finished_at` (maybe that over verbose name would be better?)

edit: I don't know why I feel like `completed` is more "final" than `finished`. Although IMO it would be odd to use `completed_at` *AND* have the `completed` status, and have rows with `completed_at` not `NULL` and `status = available`.

tomasfarias: Yeah, I agree `completed_at` feels more final and implies it was with `completed` status. We could have one timestamp per status, i.e. `completed_at`, `failed_at`, `retried_at`, but that gets redundant quickly and hard to maintain if we ever add another status.

tomasfarias (approved):
nit: I would squash migrations given that this is not yet released anywhere (I personally drop and recreate my local Postgres all the time). Not a blocker though as we can squash before moving this to production.

migrations/20231212184028_indexes.sql
tomasfarias: Happy to update indexes as we go; it's also harder to know beforehand what our hottest query patterns will be (and which we should optimize with the right index).

migrations/20231212184028_indexes.sql
bretthoerner: It could also be an array, only the final timestamp could be `completed` or `failed` anyway, anything before the final basically has to be "attempt that failed". I guess all that gets us is knowledge of when failed attempts happened, I don't know how useful that is. You're right that we could add more statuses anyway, though.

migrations/20231129172339_job_queue_table.sql
```diff
@@ -9,15 +9,20 @@ CREATE TABLE job_queue(
...
-    errors jsonb[],
+    errors jsonb [],
     max_attempts INT NOT NULL DEFAULT 1,
     finished_at TIMESTAMPTZ DEFAULT NULL,
     parameters JSONB,
-    queue TEXT NOT NULL DEFAULT 'default'::text,
+    queue TEXT NOT NULL DEFAULT 'default' :: text,
     scheduled_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
-    status job_status NOT NULL DEFAULT 'available'::job_status,
+    status job_status NOT NULL DEFAULT 'available' :: job_status,
```
bretthoerner: Ew, I use the vscode `SQL Formatter` extension and disagree with its choices, one sec.

PR #9, opened 2023-12-14
feat: Support for non transactional consumer

tomasfarias:
We do not yet want to commit to either transactional or non transactional running modes, so this makes them configurable in the consumer side. There is perhaps some opportunity for abstraction with traits, but if we commit for one of the two options early enough we could end up dropping the other code.

### DIFF
--- a/hook-consumer/src/main.rs
+++ b/hook-consumer/src/main.rs
@@ -31,7 +31,11 @@ async fn main() -> Result<(), ConsumerError> {
         config.max_concurrent_jobs,
     );
 
-    let _ = consumer.run().await;
+    if config.transactional {
+        consumer.run_tx().await?;
+    } else {
+        consumer.run().await?;
+    }
 
     Ok(())
 }
--- a/hook-consumer/src/main.rs
+++ b/hook-consumer/src/main.rs
@@ -31,11 +31,7 @@ async fn main() -> Result<(), ConsumerError> {
         config.max_concurrent_jobs,
     );
 
-    if config.transactional {
-        consumer.run_tx().await?;
-    } else {
-        consumer.run().await?;
-    }
+    consumer.run(config.transactional).await?;
 
     Ok(())
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1008,6 +1008,7 @@ dependencies = [
 name = "hook-common"
 version = "0.1.0"
 dependencies = [
+ "async-trait",
  "axum",
  "chrono",
  "http 0.2.11",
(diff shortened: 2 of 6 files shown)


hook-consumer/src/main.rs
```diff
@@ -23,15 +23,28 @@ async fn main() -> Result<(), WebhookConsumerError> {
...
-        &config.consumer_name,
-        &queue,
-        config.poll_interval.0,
-        config.request_timeout.0,
-        config.max_concurrent_jobs,
-    )?;
-
-    let _ = consumer.run().await;
+    match config.transactional {
+        true => {
```
bretthoerner: I dunno why, but `match` on a `bool` cracks me up. We're living in the future. :)

bretthoerner (approved):

hook-consumer/src/main.rs
tomasfarias: Aaah, I may be rusting too hard.

tomasfarias:
Looks like this PR has some conflicts with #4. The crux of the issue was that we updated the `WebhookJobParameters`. I originally envisioned this struct matching the interface returned by a `composeWebhook` app, which is defined in plugin-scaffold: https://github.com/PostHog/plugin-scaffold/blob/main/src/types.ts#L15.

@bretthoerner I don't think we'll get `team_id`, `plugin_config_id`, `plugin_id` as the job parameters, given that they are not part of the Webhook interface, so I'm thinking of dropping these.

EDIT: In line with this, I've nested the `WebhookJobParameters` struct under a parameters key. And added the extra parameters as top level keys. Happy to discuss this, we can also iterate further as we see how this is called from plugin server.

bretthoerner:
> @bretthoerner I don't think we'll get `team_id`, `plugin_config_id`, `plugin_id` as the job parameters, given that they are not part of the Webhook interface, so I'm thinking of dropping these.

~We should discuss, because I'm not sure where we'd get them except the from `plugin-server` POST. I agree they won't come directly from the plugin itself, but we have other metadata we'll need, and if it can't fit into the `parameters` column then we'll need a second column (which feels specific to Webhooks?) or something, and then we'd still need a type that wraps the Webhook parameters for the HTTP endpoint:~

https://github.com/PostHog/posthog/blob/fbb93232c4f4d07dce9c1c6ec6ff541a2ce3a8b9/plugin-server/src/worker/plugins/run.ts#L79-L141

edit: Ah, I hadn't read the new `WebhookPostRequestBody` when I wrote this. That seems fine to me, too, we'll just need to adjust `plugin-server` to send that shape. Although, does the full `WebhookPostRequestBody` make it into the DB parameters column? Those fields are necessary for after the job run (sending metrics/logs).

bretthoerner:
👍

PR #10, opened 2023-12-14
refactor: PgQueue no longer takes a RetryPolicy

tomasfarias:
`PgQueue` only takes a `RetryPolicy` to pass it to Jobs. In reality, only the consumer needs to know the retry policy when retrying a job. So, let's remove `RetryPolicy` from `PgQueue`, and instead take it in the consumer. Then, `retry` methods just take a duration until next retry.


bretthoerner (approved):
Nice, thanks!

PR #11, opened 2023-12-15
Append final errors to error array and treat request errors as retryable

bretthoerner:
I've been working on the janitor and noticed 2 issues:

1. We don't append the error on `fail` when it seems like we should.
2. We were treating request errors as different from `RetryableWebhookError`, but we want things like connection errors, timeouts etc to be `RetryableWebhookError` AFAIK

### DIFF
--- a/hook-consumer/src/consumer.rs
+++ b/hook-consumer/src/consumer.rs
@@ -32,7 +32,7 @@ impl<'p> WebhookConsumer<'p> {
         poll_interval: time::Duration,
         request_timeout: time::Duration,
         max_concurrent_jobs: usize,
-    ) -> Result<Self, WebhookConsumerError> {
+    ) -> Self {
         let mut headers = header::HeaderMap::new();
         headers.insert(
             header::CONTENT_TYPE,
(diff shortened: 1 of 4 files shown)


hook-consumer/src/consumer.rs
```diff
@@ -174,7 +175,11 @@ async fn send_webhook(
         .headers(headers)
         .body(body)
         .send()
-        .await?;
+        .await
+        .map_err(|e| WebhookConsumerError::RetryableWebhookError {
+            reason: e.to_string(),
```
bretthoerner, line 180: Since we surface Webhook send errors to users, I think we'll want to tame this with an enum rather than leaving it open ended, but that seems unrelated to this change and I'm leaving it for later.

tomasfarias (approved):
lgtm! Thank you

hook-consumer/src/consumer.rs
```diff
@@ -42,7 +42,8 @@ impl<'p> WebhookConsumer<'p> {
         let client = reqwest::Client::builder()
             .default_headers(headers)
             .timeout(request_timeout)
-            .build()?;
+            .build()
+            .expect("failed to construct reqwest client for webhook consumer");
```
tomasfarias, line 46: There is nothing fallible in this method if we make this change (as we'll panic), so we can return `Self` instead of a `Result`.

bretthoerner: Good eye! I feel like clippy or something should tell me that. 🤔

PR #12, opened 2023-12-18
refactor: Add a metadata field to job queue

tomasfarias:
Updates the job model to support a `metadata` field. This field is any arbitrary JSON, and can be passed to webhook producer to set plugin-server related fields (like `team_id`, or `plugin_config_id`).


migrations/20231129172339_job_queue_table.sql
```diff
@@ -11,8 +11,9 @@ CREATE TABLE job_queue(
     attempted_at TIMESTAMPTZ DEFAULT NULL,
     attempted_by TEXT[] DEFAULT ARRAY[]::TEXT[],
     created_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
-    errors jsonb[],
+    errors JSONB[],
     max_attempts INT NOT NULL DEFAULT 1,
+    metadata JSONB,
```
tomasfarias, line 16: I'm just updating the migration in place assuming this is not deployed yet.

hook-producer/src/handlers/webhook.rs
```diff
@@ -30,8 +44,22 @@ pub async fn post(
         ));
     }
 
-    let url_hostname = get_hostname(&payload.url)?;
-    let job = NewJob::new(payload.max_attempts, payload, url_hostname.as_str());
+    let url_hostname = get_hostname(&payload.parameters.url)?;
+    // We could cast to i32, but this ensures we are not wrapping.
+    let max_attempts = i32::try_from(payload.max_attempts).map_err(|_| {
```
tomasfarias, line 49: To give more context: sqlx doesn't support encoding unsigned integer types to Postgres values. However, using the type system to enforce `max_attempts` are unsigned means we would never get bugs caused by a negative max attempt, which is great as negative max attempts makes no sense. So, we take an unsigned type, and then cast it to signed for sqlx support. Using `as` for casting could wrap around if we overflow, so we make the casting with `try_from` to fail on wrap around. I don't think we'll ever need more than `i32::MAX` max attempts anyways.

Perhaps a better solution would be to have a wrapper type that implements decode/encode over a unsigned integer, but this was easier to implement and apparently good enough for now.

hook-common/src/webhook.rs
```diff
@@ -124,16 +124,13 @@ pub struct WebhookJobParameters {
...
+}
 
-    // These should be set if the Webhook is associated with a plugin `composeWebhook` invocation.
+/// `JobMetadata` required for the `WebhookConsumer` to execute a webhook.
+/// These should be set if the Webhook is associated with a plugin `composeWebhook` invocation.
+#[derive(Deserialize, Serialize, Debug, PartialEq, Clone)]
+pub struct WebhookJobMetadata {
     pub team_id: Option<i32>,
     pub plugin_id: Option<i32>,
     pub plugin_config_id: Option<i32>,
```
tomasfarias, line 135: @bretthoerner I think these should be unsigned non-optional types: Given that this struct only applies to Webhook jobs, it seems like all of these should be required metadata parameters, which we can enforce with non-optional types. I don't think there could be webhook jobs coming from anywhere but a `composeWebhook` invocation, at least not for starters, and if eventually that happens we can have a separate struct + endpoint to receive them perhaps?

Unsigned because I don't think negative ids could be generated.

Any thoughts? Wanted to focus the PR more on the metadata field so omitted updating these before checking your thoughts.

bretthoerner: Agreed, we can switch to `u32` for all of them I think.

> and if eventually that happens we can have a separate struct + endpoint to receive them perhaps?

Yeah, agree.

bretthoerner (approved):

PR #13, opened 2023-12-18
feat: Use new WebhookJobError struct for error reporting

tomasfarias:
Instead of passing errors just as strings, we should support the `app_metrics` error schema and use a struct with `type`, `details`, and `uuid` fields; a new `WebhookJobError` was added to support this schema. Consumer now passes the new struct to `fail` and `retry` calls. These methods were already implemented generically over anything that can be serialized to json, so no changes required from that side to support a struct instead of `String`.

TODO: Should the UUID have any meaning? I've included it mostly to support the `app_metrics` schema, but I'm generating a new one on every error, so it's not very useful information (at least for now).

### DIFF
--- a/hook-common/src/webhook.rs
+++ b/hook-common/src/webhook.rs
@@ -138,6 +138,7 @@ pub struct WebhookJobMetadata {
 }
 
 /// An error originating during a Webhook Job invocation.
+/// This is to be serialized to be stored as an error whenever retrying or failing a webhook job.
 #[derive(Serialize, Debug)]
 pub struct WebhookJobError {
     pub r#type: app_metrics::ErrorType,
@@ -146,6 +147,9 @@ pub struct WebhookJobError {
     pub uuid: uuid::Uuid,
 }
 
+/// Webhook jobs boil down to an HTTP request, so it's useful to have a way to convert from &reqwest::Error.
+/// For the convertion we check all possible error types with the associated is_* methods provided by reqwest.
+/// Some precision may be lost as our app_metrics::ErrorType does not support the same number of variants.
 impl From<&reqwest::Error> for WebhookJobError {
     fn from(error: &reqwest::Error) -> Self {
         if error.is_body() || error.is_decode() {
--- a/hook-common/src/webhook.rs
+++ b/hook-common/src/webhook.rs
@@ -1,9 +1,11 @@
 use std::collections;
+use std::convert::From;
 use std::fmt;
 use std::str::FromStr;
 
 use serde::{de::Visitor, Deserialize, Serialize};
 
+use crate::kafka_messages::{app_metrics, serialize_uuid};
 use crate::pgqueue::PgQueueError;
 
 /// Supported HTTP methods for webhooks.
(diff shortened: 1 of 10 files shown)


hook-common/src/webhook.rs
```diff
@@ -134,3 +134,22 @@ pub struct WebhookJobMetadata {
...
     pub plugin_config_id: Option<i32>,
 }
+
+/// An error originating during a Webhook Job invocation.
+#[derive(Deserialize, Serialize, Debug, PartialEq, Clone)]
+pub struct WebhookJobError {
+    pub r#type: String,
+    pub details: String,
+    #[serde(with = "uuid::serde::compact")]
+    pub uuid: uuid::Uuid,
```
tomasfarias: Perhaps we should serialize this as a string...

tomasfarias: Need to add some tests to assert the behavior.

bretthoerner: Here's the `AppMetrics` types I made: https://github.com/PostHog/rusty-hook/pull/14/files#diff-a00f575cc51a6d6428477d58551304b1b9fefb596f6e9dd39f3e34870773e028

hook-common/src/webhook.rs
```diff
@@ -134,3 +134,22 @@ pub struct WebhookJobMetadata {
     pub plugin_id: Option<i32>,
     pub plugin_config_id: Option<i32>,
 }
+
+/// An error originating during a Webhook Job invocation.
+#[derive(Deserialize, Serialize, Debug, PartialEq, Clone)]
+pub struct WebhookJobError {
+    pub r#type: String,
```
bretthoerner: TIL `r#type`, I had gone with:

```
    #[serde(rename = "type", serialize_with = "serialize_type")]
    pub type_: PluginLogEntryType,
```

Do you have to call it `r#type` when you use it, too?

hook-common/src/webhook.rs
```diff
@@ -134,3 +134,22 @@ pub struct WebhookJobMetadata {
...
+    pub details: String,
```
bretthoerner: `error_details` actually has an expected shape and fields that are used for things, if you want to copy from my Kafka message types.

bretthoerner (commented):
> Should the UUID have any meaning?

I couldn't find any meaning in `plugin-server`, I think new UUID7 is fine/expected.

hook-common/src/webhook.rs
tomasfarias: Yeah, it's required everywhere, e.g.:
```
    let r#type = "My type".to_string();
    let error = WebhookJobError {
        r#type,    
    };
    println!("{}", error.r#type);
```

Otherwise the compiler will scream at you that type is a keyword.

I used it as an alternative to `rename`-ing an alias, as:
```
    let json_str = serde_json::to_string(&error).unwrap();
    println!("{}", json_str);
```

Outputs:
```
{"type":"My type"}
```

hook-common/src/webhook.rs
tomasfarias: Interesting.

I was looking at the [ddl query](https://github.com/PostHog/posthog/blob/35cd4656b91d705215c9f3366e6a5509774318f6/posthog/models/app_metrics/sql.py#L15) for `app_metrics` when writing this. Looks like everything in details is dumped into a string in Clickhouse. In other words, plugin-server is the one ensuring the schema of `error_details` matches whatever is required, Clickhouse just stores and returns a string.

hook-common/src/webhook.rs
tomasfarias: I do not have a strong preference for one or the other by the way, happy if we agree on an alias otherwise; `ty` is a common one for `type`, and `type_` is fine too. `r#type` just came out first when writing this.

bretthoerner: I don't care at all, I just hadn't seen it (or certainly I forgot it).

hook-common/src/webhook.rs
bretthoerner: Yeah, that's true, what I mean is that [parts of the frontend look for those specific fields and use them](https://github.com/PostHog/posthog/blob/e3b2c9f9a1e9d7481f23796cbc3ede495e62e481/frontend/src/scenes/apps/ErrorDetailsModal.tsx#L60-L90). That felt like a good use of `Option<String>` to me so that callers may know it's expected and provide it if possible. But it's true that CH doesn't care.

hook-common/src/webhook.rs
```diff
@@ -134,3 +136,99 @@ pub struct WebhookJobMetadata {
...
+            )
+        } else if error.is_connect()
+            || error.is_builder()
+            || error.is_request()
+            || error.is_redirect()
+        {
+            // Builder errors seem to be related to unable to setup TLS, so I'm bundling them in connection.
+            WebhookJobError::new_connection(&error.to_string())
+        } else {
+            // We can't match on Kind as some types do not have an associated variant in Kind (e.g. Timeout).
```
bretthoerner: Interesting. I think this this is fine, but could you also check for `is_timeout()` up top and short-circuit, and then match on Kind?

I'm not sure if it would be cleaner or not, up to you.

bretthoerner (approved):

hook-common/src/webhook.rs
tomasfarias: Unfortunately, the `kind` field of the `inner` struct held by `reqwest::Error` is private, so we can't match on it anyways :sweat: 

I'm not entirely sure about what each error kind means so we may need to adjust these as we go along.

bretthoerner: Damn, it's a shame that a reqwest upgrade could lead to a crash in prod (right? new error type?).

Hopefully that's unlikely. I guess the only other option is to have a catch-all type and then log it loudly somewhere we'd see it. (Not saying we should do this, just pondering.)

tomasfarias: Yeah, a new error type would cause us to panic (assuming such an error is returned somewhere). There is also a risk of bugs in reqwest causing us to panic: Most of the `is_*` methods check for `Kind::{method_name}` to return a `bool`, but some of them do more complicated checks, for example to determine whether an error `is_timeout`. A bug that causes all of the `is_*` methods to return `false` could also cause us to panic.

The catch-all type would work, maybe `Unknown` or something.

tomasfarias: Alternatively, we could ignore whatever reqwest says the error is and determine the error type based on the call that triggered it, for example:
1. `send()` can only be a `Timeout` or `Connection`:
```
if error.is_timeout() {
    ErrorType::Timeout
} else {
    // Ignore everything else, use connection as catch-all because it happened in send()
    ErrorType::Connection
}
```
2. Parsing URL/headers methods can only be a `Parse` error.
3. After receiving a response, error can only be `HttpStatus`.

2 and 3 are fairly straight-forward I think, but 1 is the one with troubles: some error types like `is_builder`, `is_body`, or `is_decode`, may be grouped together with `Connection`, even though they are not really connection errors. Unfortunately, reqwest docs don't provide examples of what can trigger each error type.

tomasfarias: I kind of like that approach more, just to avoid the production panic: I rather have a broad definition of "Connection Error" than crash. Instead of having to add an error type to fix a crash, we can add more error types when needed for more specific reporting.

bretthoerner: Sounds good to me!

PR #14, opened 2023-12-18
Add hook-janitor skeleton and Kafka message types

bretthoerner:
I'm filling in the actual SQL now, but the skeleton is ready and I wanted to show the Kafka types.


bretthoerner:
Blah, clippy is mad about my unused fields/imports, but nothing else serious. They'll be used when I finish the SQL parts.

hook-common/src/kafka_messages/app_metrics.rs
```diff
@@ -0,0 +1,123 @@
...
+use super::{serialize_datetime, serialize_uuid};
+
+#[derive(Serialize)]
+pub enum AppMetricCategory {
+    ProcessEvent,
+    OnEvent,
+    ScheduledTask,
+    Webhook,
+    ComposeWebhook,
+}
```
tomasfarias, line 14: I get that we are supporting all possible categories, but given that we would only ever output `Webhook` metrics, we could just have that as a const string?

bretthoerner: Sure, I don't mind that. My thinking is that we'll want a more general Rust library at some point (as in, I wish this type lived in some `posthog-rust-lib` repo instead), and I wanted my "port" of types to be faithful so I didn't miss anything.

tomasfarias: Yeah, I see. My original comment just imagined a lot of unnecessary `match` statements for only one enum variant. Do you think we should move these types to capture-rs and use that as a library as we discussed? I'll need to reference them from consumer, so wondering if you feel better with this being here in `hook-common` or in capture.

hook-common/src/kafka_messages/app_metrics.rs
```diff
@@ -0,0 +1,123 @@
...
+    Connection,
+    HttpStatus(u16),
+}
+
+#[derive(Serialize)]
+pub struct ErrorDetails {
+    pub error: Error,
+    // TODO: The plugin-server sends the entire raw event with errors. In order to do this, we'll
+    // have to pass the entire event when we enqueue items, and store it in the Parameters JSONB
+    // column. We should see if it's possible to work around this before we commit to it.
```
tomasfarias, line 28: As we discussed, we are only sticking with an error message as part of the details, so I wouldn't include event as a field.

hook-common/src/kafka_messages/plugin_logs.rs
```diff
@@ -0,0 +1,130 @@
...
+pub struct PluginLogEntry {
+    #[serde(serialize_with = "serialize_source")]
+    pub source: PluginLogEntrySource,
+    #[serde(rename = "type", serialize_with = "serialize_type")]
+    pub type_: PluginLogEntryType,
+    #[serde(serialize_with = "serialize_uuid")]
+    pub id: Uuid,
+    pub team_id: u32,
+    pub plugin_id: u32,
+    pub plugin_config_id: u32,
```
tomasfarias, line 35: The main reason we developed the generic `log_entries` was because of these fields: not every system that wants to write logs has a numerical id associated with it, some use arbitrary ids, like UUID, and generating one could risk colliding with existing plugins.

However, like we discussed, for webhooks we will have an associated plugin, so using plugin_log_entries as our log table is possible.

hook-common/src/kafka_messages/app_metrics.rs
bretthoerner: I forgot about this comment, I do think we should probably add the event ID here (which means adding it to `metadata` I think), so we have some way to tie it back to the event.

hook-common/src/kafka_messages/app_metrics.rs
bretthoerner: Since capture doesn't need these, I feel like leaving them here is fine and we'll hopefully be able to combine repos at the offsite.

hook-common/src/kafka_messages/app_metrics.rs
bretthoerner: It doesn't have to be now or anything, though.

tomasfarias: Yeah, metadata can be extended with `event_id` :+1:

hook-janitor/src/handlers/app.rs
```diff
@@ -0,0 +1,21 @@
+use axum::{routing, Router};
+use metrics_exporter_prometheus::PrometheusHandle;
+
+use hook_common::metrics;
+
+pub fn app(metrics: Option<PrometheusHandle>) -> Router {
+    Router::new()
```
tomasfarias, line 7: I'm confused by this http server. Who's going to be calling the janitor? Can't the janitor just track it's own metrics while running? Why does it need a server?

hook-janitor/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+
+async fn listen(app: Router, bind: String) -> Result<()> {
+    let listener = tokio::net::TcpListener::bind(bind).await?;
+
+    axum::serve(listener, app).await?;
+
+    Ok(())
+}
+
+async fn cleanup_loop(cleaner: Box<dyn Cleaner>, interval_secs: u64) -> Result<()> {
```
tomasfarias: nit: I'd avoid returning a result if there's nothing here that can return an error, otherwise we force callers to handle an error that will never happen. But it may be that this signature is to accommodate `cleanup` returning errors in the future?

hook-janitor/src/handlers/app.rs
bretthoerner: Hmm, did I miss an easier way to expose the Prometheus HTTP metrics endpoint (that will be scraped by Victoria Metrics)?

hook-janitor/src/main.rs
bretthoerner: Nope, I think this was from an earlier iteration where I was having trouble making the `select` happy, but I can remove this now for sure.

hook-janitor/src/main.rs
```diff
@@ -0,0 +1,86 @@
...
+    let semaphore = Semaphore::new(1);
+    let mut interval = tokio::time::interval(Duration::from_secs(interval_secs));
+
+    loop {
+        let _permit = semaphore.acquire().await;
+        interval.tick().await;
+        cleaner.cleanup().await;
```
tomasfarias, line 35: Should we spawn a task here? Although I don't know if we need concurrency; in fact multiple cleaners running at the same time could be a problem.

hook-janitor/src/handlers/app.rs
tomasfarias: Ahh, is to scrap metrics. I guess I didn't like seeing pretty much the same server code as in producer. Maybe we can abstract this somehow? I will have to add the same thing in consumer probably

hook-janitor/src/main.rs
```diff
@@ -0,0 +1,85 @@
...
+                .expect("unable to create webhook cleaner"),
+            )
+        }
+    };
+
+    let cleanup_loop = Box::pin(cleanup_loop(cleaner, config.cleanup_interval_secs));
+
+    let recorder_handle = metrics::setup_metrics_recorder();
+    let app = handlers::app(Some(recorder_handle));
+    let http_server = Box::pin(listen(app, config.bind()));
```
tomasfarias, line 74: Alternatively, we could `tokio::spawn` a task to run this. Otherwise the default multi-threaded runtime isn't getting much action from our end, doesn't it?

tomasfarias (approved):
Overall, looks good to me, some dead code, but it will get used as we develop further.

hook-janitor/src/main.rs
bretthoerner: Yeah, we want to block and complete. Let's say we run every 60 seconds and for some reason this starts taking 70 seconds, I don't want them to overlap.

hook-janitor/src/main.rs
bretthoerner: I'm not sure I follow, why do we need/want to `spawn` this?

hook-janitor/src/handlers/app.rs
bretthoerner: Yeah, I think especially as part of a unified Rust workspace we should have a nice server wrapper that sets up metrics and everything.

hook-janitor/src/main.rs
bretthoerner: My understanding is that the `select` here means both will run concurrently: https://github.com/PostHog/rusty-hook/blob/5a5c3bed8c3f6c78bbe869b8a83cbd288cd90eb2/hook-janitor/src/main.rs#L76

And if I run the server, I can `curl` metrics -- it doesn't seem possible for that to work if these `await`s were blocking everything?: https://github.com/PostHog/rusty-hook/blob/5a5c3bed8c3f6c78bbe869b8a83cbd288cd90eb2/hook-janitor/src/main.rs#L33-L36

PR #15, opened 2023-12-19
Implement Webhook cleanup

bretthoerner:


hook-common/src/webhook.rs
```diff
@@ -139,12 +139,10 @@ pub struct WebhookJobMetadata {
 
 /// An error originating during a Webhook Job invocation.
 /// This is to be serialized to be stored as an error whenever retrying or failing a webhook job.
-#[derive(Serialize, Debug)]
+#[derive(Deserialize, Serialize, Debug)]
 pub struct WebhookJobError {
     pub r#type: app_metrics::ErrorType,
     pub details: app_metrics::ErrorDetails,
-    #[serde(serialize_with = "serialize_uuid")]
-    pub uuid: uuid::Uuid,
```
bretthoerner, line 147: I dropped this because it makes `error` columns unnecessarily unique which hurts the `GROUP BY` aggregation I do. I don't think we get any special value of these `UUID`s and so I'd rather generate them when I create the `AppMetric`.

bretthoerner:
I had an annoyingly long detour in testing (see TODO in the code), I still need to wrap up the tests tomorrow.

hook-common/src/pgqueue.rs
```diff
@@ -827,18 +844,18 @@ mod tests {
         "https://myhost/endpoint".to_owned()
     }
 
-    #[tokio::test]
-    async fn test_can_dequeue_job() {
+    #[sqlx::test(migrations = "../migrations")]
```
bretthoerner, line 781: This is compatible with `#[tokio::test]` but it gives us a DB per test.

We shouldn't have to provide `migrations` but it was giving me an error without it, maybe something to do with using workspaces? It didn't seem worth investigating right now.

hook-janitor/src/webhooks.rs
```diff
@@ -1,55 +1,454 @@
...
+    async fn get_failed_rows(&self, tx: &mut Transaction<'_, Postgres>) -> Result<Vec<FailedRow>> {
+        let base_query = format!(
+            r#"
+            SELECT DATE_TRUNC('hour', finished_at) AS hour,
+                   (metadata->>'team_id')::bigint AS team_id,
+                   (metadata->>'plugin_config_id')::bigint AS plugin_config_id,
+                   CASE
+                       WHEN array_length(errors, 1) > 1
+                       THEN errors[array_length(errors, 1)]
+                       ELSE errors[1]
```
tomasfarias: Not entirely sure why we need the check, I may be missing something, wouldn't `errors[array_upper(errors, 1)]` be enough?

hook-janitor/src/webhooks.rs
```diff
@@ -1,55 +1,454 @@
...
+            .bind(&self.queue_name)
+            .fetch_all(&mut **tx)
+            .await
+            .map_err(|e| WebhookCleanerError::GetFailedRowsError { error: e })?;
+
+        Ok(rows)
+    }
+
+    fn serialize_failed_rows(&self, failed_rows: Vec<FailedRow>) -> Result<Vec<String>> {
+        let mut payloads = Vec::new();
```
tomasfarias: This method doesn't seem to access `self`, so I would personally opt to implement `Serialize` on `FailedRow` instead. Or perhaps `Into<AppMetric> for FailedRow` and then `Serialize` on `AppMetric`.

tomasfarias: Same thing with `serialize_completed_rows`.

hook-janitor/src/webhooks.rs
```diff
@@ -1,55 +1,454 @@
...
+                }
+            }
+        }
+
+        Ok(())
+    }
+
+    async fn delete_observed_rows(&self, tx: &mut Transaction<'_, Postgres>) -> Result<u64> {
+        // This DELETE is only safe because we are in serializable isolation mode, see the note
+        // in `start_serializable_txn`.
```
tomasfarias, line 294: This works, but I wish there was a way to enforce this: Seems like we take a tx, but that can be any transaction, not necessarily one that we started with serializable isolation mode by calling `start_serializable_txn`. Maybe we could use a wrapper type?

hook-janitor/src/webhooks.rs
```diff
@@ -1,55 +1,454 @@
...
+        let completed_rows = self.get_completed_rows(&mut tx).await?;
+        let completed_agg_row_count = completed_rows.len();
+        let completed_kafka_payloads = self.serialize_completed_rows(completed_rows)?;
+
+        let failed_rows = self.get_failed_rows(&mut tx).await?;
+        let failed_agg_row_count = failed_rows.len();
+        let mut failed_kafka_payloads = self.serialize_failed_rows(failed_rows)?;
+
+        let mut all_kafka_payloads = completed_kafka_payloads;
+        all_kafka_payloads.append(&mut failed_kafka_payloads);
```
tomasfarias: I'd prefer we use `extend` here. `extend` consumes the `failed_kafka_payloads` vec, so we won't be able to access it later. By using append, we are still keeping ownership of `failed_kafka_payloads`, opening the door up for bugs. Not seeing any bugs right now, but just feel like it's safer.

hook-janitor/src/webhooks.rs
```diff
@@ -1,55 +1,454 @@
...
+        // TODO: I spent a lot of time trying to get the mock Kafka consumer to work, but I think
+        // I've identified an issue with the rust-rdkafka library:
+        //   https://github.com/fede1024/rust-rdkafka/issues/629#issuecomment-1863555417
+        //
+        // I wanted to test the messages put on the AppMetrics topic, but I think we need to figure
+        // out that issue about in order to do so. (Capture uses the MockProducer but not a
+        // Consumer, fwiw.)
+        //
+        // For now, I'll probably have to make `cleanup_impl` return the row information so at
+        // least we can inspect that for correctness.
```
tomasfarias: I don't know if it would be more work, but we can always spin-up a kafka in the docker compose stack, so no mocking is required.

hook-janitor/src/webhooks.rs
```diff
@@ -1,55 +1,454 @@
...
+            .begin()
+            .await
+            .map_err(|e| WebhookCleanerError::StartTxnError { error: e })?;
+
+        // We use serializable isolation so that we observe a snapshot of the DB at the time we
+        // start the cleanup process. This prevents us from accidentally deleting rows that are
+        // added (or become 'completed' or 'failed') after we start the cleanup process.
+        //
+        // If we find that this has a significant performance impact, we could instead move
+        // rows to a temporary table for processing and then deletion.
```
tomasfarias, line 189: For normal webhook processing, we are skipping all the locked rows (for two reasons, we run with `SKIP LOCKED` and we check for status not in (`failed`, `completed`). So, I don't expect major performance impact there. If this transaction stays open for too long, that could delay auto-vacuum, and eventually degrade overall performance. However, I don't see this running for that long, and as long as we timeout effectively on Postgres side we should catch any lingering open transactions.

hook-janitor/src/webhooks.rs
```diff
@@ -1,55 +1,454 @@
...
+        Ok(tx)
+    }
+
+    async fn get_completed_rows(
+        &self,
+        tx: &mut Transaction<'_, Postgres>,
+    ) -> Result<Vec<CompletedRow>> {
+        let base_query = format!(
+            r#"
+            SELECT DATE_TRUNC('hour', finished_at) AS hour,
```
tomasfarias, line 201: Doesn't this mean that if we ever change the aggregation on app metrics' side, we will be losing information here? For example, if we ever start trunc-ing by 15 minutes on app metrics, we will have already sent truncated dates that provide no precision for 15 minutes aggregates.

tomasfarias: Ugh, but I can't think of a better alternative, so I think we just live with this, at least for now.

hook-common/src/pgqueue.rs
tomasfarias: We can take a look later; I wasn't familiar with `sqlx::test`, good suggestion!

tomasfarias: It does mean we need to have the database url set in the environment (I see the provided `.env`), so perhaps we could document that in the README

hook-common/src/webhook.rs
tomasfarias: Yup, I asked this question in my PR, happy to see it dropped, thank you :+1:

hook-janitor/src/webhooks.rs
tomasfarias: This is a bit nitpicky, especially for a first implementation of cleanup when the risk of mixing up transactions is very low. Happy to leave this for later, although I am a big fan of checking stuff at compile time if possible.

hook-janitor/src/webhooks.rs
bretthoerner: You're right, that's much cleaner, I missed that function.

hook-janitor/src/webhooks.rs
bretthoerner: Good idea!

hook-janitor/src/webhooks.rs
bretthoerner: No, I think this is a good idea, and should be easy enough by just wrapping it in a simple struct for some protection.

hook-janitor/src/webhooks.rs
bretthoerner: Ah, I missed that we weren't giving up ownership with `append`, thanks.

hook-janitor/src/webhooks.rs
bretthoerner: That is true, but it seems like a big thing to change the aggregation level of the `app_metrics` table. I would think we'd make a new topic and revisit producers to it, but I guess that's a big assumption I'm making.

I don't have to do any aggregation here, but it seems like a shame to not, since the _only_ reason we're reading the rows is to write to `app_metrics`.

How about I PR to PostHog to document that the aggregation is import to existing producers like rusty-hook?

hook-janitor/src/webhooks.rs
bretthoerner: That's true, I may check that out today.

hook-janitor/src/webhooks.rs
bretthoerner: https://github.com/PostHog/posthog/pull/19449

hook-janitor/src/webhooks.rs
```diff
@@ -84,6 +84,10 @@ struct FailedRow {
     failures: u32,
 }
 
+// A simple wrapper type that ensures we don't use any old Transaction object when we need one
+// that has set the isolation level to serializable.
+struct SerializableTxn<'a>(Transaction<'a, Postgres>);
```
tomasfarias, line 125: Nice, thank you :+1:

hook-janitor/src/webhooks.rs
tomasfarias: I've made a comment on that PR, thank you for opening it.

Regardless, I'm okay with going forward with aggregation. Worst possible case, if we end up showing metrics with more timestamp precision, we can change this later and the only consequence is our first set of metrics will look perfectly on the hour mark, which is not a big deal.

hook-janitor/src/webhooks.rs
bretthoerner: Figured out the mocking, needed `.set("auto.offset.reset", "earliest")`

bretthoerner:
I got Kafka mocking working and finished up the tests, so this is ready (again).

hook-common/src/kafka_messages/app_metrics.rs
```diff
@@ -77,19 +88,88 @@ where
...
-        ErrorType::HttpStatus(s) => format!("HTTP Status: {}", s),
-        ErrorType::Parse => "Parse Error".to_owned(),
+        Some(error_type) => error_type,
+        None => return serializer.serialize_none(),
+    };
+
+    let error_type = match error_type {
+        ErrorType::ConnectionError => "Connection Error".to_owned(),
+        ErrorType::TimeoutError => "Timeout".to_owned(),
+        ErrorType::BadHttpStatus(s) => format!("HTTP Status: {}", s),
```
tomasfarias: nit: I don't know if these names matter too much, but for consistency:

```suggestion
        ErrorType::TimeoutError => "Timeout Error".to_owned(),
        ErrorType::BadHttpStatus(s) => format!("Bad HTTP Status: {}", s),
```

hook-common/src/kafka_messages/app_metrics.rs
```diff
@@ -77,19 +88,88 @@ where
...
+                    "composeWebhook",
+                ],
+            ))
+        }
+    };
+
+    Ok(category)
+}
+
+fn serialize_error_type<S>(error_type: &Option<ErrorType>, serializer: S) -> Result<S::Ok, S::Error>
```
tomasfarias, line 120: nit: I may be wrong but I think there is a Serialize/Deserialize implementation for `Option` and `&Option`, so we could implement Serialize/Deserialize on `ErrorType` instead and we would get Serialize/Deserialize on `&Option<ErrorType>` too. Not a big deal, but would save the check for `Some(error_type)`/`None`

hook-common/src/kafka_messages/app_metrics.rs
```diff
@@ -77,19 +88,88 @@ where
...
+    let opt = Option::<String>::deserialize(deserializer)?;
+    let error_type = match opt {
+        Some(s) => {
+            let error_type = match &s[..] {
+                "Connection Error" => ErrorType::ConnectionError,
+                "Timeout" => ErrorType::TimeoutError,
+                _ if s.starts_with("HTTP Status:") => {
+                    let status = &s["HTTP Status:".len()..];
+                    ErrorType::BadHttpStatus(status.parse().map_err(serde::de::Error::custom)?)
+                }
```
tomasfarias: Also for consistency, since the `ConnectionError` and `ParseError` variants include their full name in the str:
```suggestion
                "Timeout Error" => ErrorType::TimeoutError,
                _ if s.starts_with("Bad HTTP Status:") => {
                    let status = &s["Bad HTTP Status:".len()..];
                    ErrorType::BadHttpStatus(status.parse().map_err(serde::de::Error::custom)?)
                }
```

hook-common/src/kafka_messages/app_metrics.rs
```diff
@@ -77,19 +88,88 @@ where
...
+                "Parse Error" => ErrorType::ParseError,
+                _ => {
+                    return Err(serde::de::Error::unknown_variant(
+                        &s,
+                        &[
+                            "Connection Error",
+                            "Timeout",
+                            "HTTP Status: <status>",
+                            "Parse Error",
```
tomasfarias: Same, just for consistency, but not a big deal:
```suggestion
                            "Connection Error",
                            "Timeout Error",
                            "Bad HTTP Status: <status>",
                            "Parse Error",
```

hook-janitor/src/webhooks.rs
```diff
@@ -1,55 +1,761 @@
...
+    use rdkafka::mocking::MockCluster;
+    use rdkafka::producer::{DefaultProducerContext, FutureProducer};
+    use rdkafka::{ClientConfig, Message};
+    use sqlx::{PgPool, Row};
+    use std::collections::HashMap;
+    use std::str::FromStr;
+
+    const APP_METRICS_TOPIC: &str = "app_metrics";
+
+    async fn create_mock_kafka() -> (
```
tomasfarias, line 409: Nice! We may want to re-use this in other tests, so I'm keeping it in mind.

hook-janitor/src/webhooks.rs
```diff
@@ -1,55 +1,761 @@
...
+        // We haven't committed, so the rows are still visible from outside the txn.
+        assert_eq!(get_count_from_new_conn(&db, "completed").await, 7);
+        assert_eq!(get_count_from_new_conn(&db, "available").await, 1);
+
+        webhook_cleaner.commit_txn(tx).await.unwrap();
+
+        // We have committed, what remains are:
+        // * The 1 available job we completed while the txn was open.
+        // * The 2 brand new jobs we added while the txn was open.
+        // * The 1 running job that didn't change.
```
tomasfarias, line 754: Nice!

tomasfarias (approved):
Looks pretty good to me, some minor nitty comments here and there that do not necessarily need addressing right now. Thanks!

hook-common/src/kafka_messages/app_metrics.rs
bretthoerner: Good catch. It's good to keep these clean and clear because they get (eventually) exposed to the user.

hook-common/src/kafka_messages/app_metrics.rs
bretthoerner: Hmmm, maybe I need a special serde annotation to make it work?

```
error[E0308]: mismatched types
   --> hook-common/src/kafka_messages/app_metrics.rs:46:23
    |
46  | #[derive(Deserialize, Serialize, Debug, PartialEq, Clone)]
    |                       ^^^^^^^^^ expected `ErrorType`, found `&Option<ErrorType>`
...
67  |         serialize_with = "serialize_error_type",
    |                          ---------------------- arguments to this function are incorrect
    |
    = note:   expected enum `ErrorType`
            found reference `&'__a std::option::Option<ErrorType>`
note: function defined here
   --> hook-common/src/kafka_messages/app_metrics.rs:120:4
    |
120 | fn serialize_error_type<S>(error_type: ErrorType, serializer: S) -> Result<S::Ok, S::Error>
    |    ^^^^^^^^^^^^^^^^^^^^    ---------------------
    = note: this error originates in the derive macro `Serialize` (in Nightly builds, run with -Z macro-backtrace for more info)
```

bretthoerner: I did a little searching but I don't know if this is possible, if you know a way I'm happy to come back and clean it up.

PR #18, opened 2023-12-21
feat: Serve basic metrics on consumer

tomasfarias:
Serve an endpoint to scrap Prometheus metrics from hook-consumer. This PR added a `serve_metrics` function to the common library as it was getting repetitive to have to start a new tcplistener and axum server every time. However, I have not replaced the axum servers started for janitor and producer, in case I am missing something. Potentially, this function can be hidden behind a feature if it conflicts with others.

Also notably, we are starting the axum server as a separate Tokio task. Some graceful shutdown handling may be required, but since that's missing in the rest of the consumer, it is left as TODO in a follow-up PR.

The metrics provided so far are:
* webhook_jobs_total: A count of total jobs processed.
* webhook_jobs_completed: A count of jobs finished successfully.
* webhook_jobs_failed: A count of jobs finished unsuccessfully.
* webhook_jobs_retried: A count of jobs scheduled for retry.
* webhook_jobs_processing_duration_seconds: Total duration in seconds to process a job.

Metrics are labelled by queue and target, but perhaps other labels should be included? Open to suggestions!

### DIFF
--- a/hook-consumer/src/consumer.rs
+++ b/hook-consumer/src/consumer.rs
@@ -19,6 +19,8 @@ trait WebhookJob: PgQueueJob + std::marker::Send {
     fn parameters(&self) -> &WebhookJobParameters;
     fn metadata(&self) -> &WebhookJobMetadata;
     fn attempt(&self) -> i32;
+    fn queue(&self) -> String;
+    fn target(&self) -> String;
 }
 
 impl WebhookJob for PgTransactionJob<'_, WebhookJobParameters, WebhookJobMetadata> {
@@ -33,6 +35,14 @@ impl WebhookJob for PgTransactionJob<'_, WebhookJobParameters, WebhookJobMetadat
     fn attempt(&self) -> i32 {
         self.job.attempt
     }
+
+    fn queue(&self) -> String {
+        self.job.queue.to_owned()
+    }
+
+    fn target(&self) -> String {
+        self.job.target.to_owned()
+    }
 }
 
 impl WebhookJob for PgJob<WebhookJobParameters, WebhookJobMetadata> {
@@ -46,6 +56,14 @@ impl WebhookJob for PgJob<WebhookJobParameters, WebhookJobMetadata> {
 
     fn attempt(&self) -> i32 {
         self.job.attempt
+    }
+
+    fn queue(&self) -> String {
+        self.job.queue.to_owned()
+    }
+
+    fn target(&self) -> String {
+        self.job.target.to_owned()
     }
 }
 
@@ -170,6 +188,13 @@ async fn spawn_webhook_job_processing_task<W: WebhookJob + 'static>(
         .await
         .expect("semaphore has been closed");
 
+    let labels = [
+        ("queue", webhook_job.queue()),
+        ("target", webhook_job.target()),
+    ];
+
+    metrics::increment_counter!("webhook_jobs_total", &labels);
+
     tokio::spawn(async move {
         let result = process_webhook_job(client, webhook_job, &retry_policy).await;
         drop(permit);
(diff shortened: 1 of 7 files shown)


hook-consumer/src/consumer.rs
```diff
@@ -170,9 +185,21 @@ async fn spawn_webhook_job_processing_task<W: WebhookJob + 'static>(
...
+
+    metrics::increment_counter!("webhook_jobs_total", &labels);
+
     tokio::spawn(async move {
+        let now = tokio::time::Instant::now();
         let result = process_webhook_job(client, webhook_job, &retry_policy).await;
         drop(permit);
+
+        let elapsed = now.elapsed().as_secs_f64();
+        metrics::histogram!("webhook_jobs_processing_duration_seconds", elapsed, &labels);
```
bretthoerner: I wonder if this should only be called for non-timeouts? It seems like timeouts would pollute the metric since they would be truncated anyway, they could have their own metric (just a count)?

bretthoerner (approved):
> I have not replaced the axum servers started for janitor and producer

It's a relatively minor opinion, but it's nice when everything is available on one port. It saves you from having to fish for which port to hit for metrics vs other things. I'm only mentioning because it would be nice if the Axum server could be reused for other things. (Even if it doesn't seem like the service needs HTTP. It might for example, might need to expose some kind of `/debug` endpoint in the future, or it might expose a memory/cpu sampling endpoint, etc. Actually, that's an example of something that'd be kinda nice to add to every service just like `/metrics`.)

Could help with shutdown stuff etc like you mentioned, too.

Like, could our common library do `Router::new().route("/metrics", ...)` and return the Router so the caller can then add its own pieces?

Anyway, not critical, just thinking out loud.

tomasfarias:
> Like, could our common library do Router::new().route("/metrics", ...) and return the Router so the caller can then add its own pieces?

Yeah, that's a nice compromise IMO :+1: I can change the common library a bit to separate router creation in case any caller needs to add it's own routes.

hook-consumer/src/consumer.rs
tomasfarias: Good point, we can have another counter for `webhook_jobs_timedout`.

tomasfarias: Alternatively, we could add an error label to the `webhook_jobs_failed` metric.

tomasfarias:
> but it's nice when everything is available on one port. It saves you from having to fish for which port to hit for metrics vs other things.

I set the consumer on a different port by default on purpose: I imagined when testing locally you will want to spawn all producer, consumer, and janitor. Having all of them serving the same port would cause conflicts so I bumped consumer to 8001. I see the point that in production this is not a concern though, and the reverse is true: it's more convenient to have everything under the same port. Wish there were test vs production settings...

hook-consumer/src/consumer.rs
tomasfarias: This made me think that wouldn't tracking duration seconds on quick failures (like parsing errors) skew the metric lower? Kind of the opposite of tracking it for timeouts. I'm thinking of putting this histogram metric only on successful processing, and maybe have a separate one for failures?

bretthoerner (approved):
Looks great. I was thinking we may want a label detailing _why_ things failed, but we can wait and see if that's true. It might be not be that useful as an aggregate across all `target`s.

I wonder where we'll get insight like that? If Grafana could query the `job_queue` itself it could at be possible to show number of currently failing `target`s and their reason (for the current point in time only), I guess? Keeping possibly high cardinality info like that over time is a job for CH and not Prometheus, really.

Anyway, I'm rambling. ☕

tomasfarias:
> Looks great. I was thinking we may want a label detailing why things failed, but we can wait and see if that's true. It might be not be that useful as an aggregate across all targets.

Yeah, I was thinking about sticking `error_type` as a label of `webhook_jobs_failed`, that should stay low cardinality (I don't foresee us going crazy adding variants to that enum), and we at least know if it was a timeout, connection, or bad http status. 

But yeah, we can also stick that in later.

bretthoerner:
Right, it just dawned on me that knowing that is kind of useless without also knowing the `target`, which is probably too high cardinality.

PR #19, opened 2023-12-28
feat: Implement support for retrying jobs to different queue

tomasfarias:
This PR implements separate queue retrying. `retry` methods now take a `retry_queue` which can be set to a different queue to set when updating the job. The `retry_queue` is provided by the `RetryPolicy`, which currently only returns a configured value if set. If nothing is set then `retry_queue` will just be the original job queue.

Also in this PR:
* Added tests for retry.rs.
* Fix some bugs in retry.rs.

### DIFF
--- a/hook-common/src/pgqueue.rs
+++ b/hook-common/src/pgqueue.rs
@@ -109,28 +109,51 @@ impl<J, M> Job<J, M> {
         self.attempt >= self.max_attempts
     }
 
-    /// Consume Job to retry it.
-    /// This returns a RetryableJob that can be enqueued by PgQueue.
-    ///
-    /// # Arguments
-    ///
-    /// * `error`: Any JSON-serializable value to be stored as an error.
-    fn retry<E: serde::Serialize>(self, error: E) -> RetryableJob<E> {
+    fn as_retryable(self) -> RetryableJob {
         RetryableJob {
             id: self.id,
             attempt: self.attempt,
-            error: sqlx::types::Json(error),
             queue: self.queue,
+            retry_queue: None,
         }
     }
 
     /// Consume Job to complete it.
     /// This returns a CompletedJob that can be marked as completed by PgQueue.
-    fn complete(self) -> CompletedJob {
-        CompletedJob {
+    async fn complete<'c, E>(
+        self,
+        queue_table: &str,
+        executor: E,
+    ) -> Result<CompletedJob, sqlx::Error>
+    where
+        E: sqlx::Executor<'c, Database = sqlx::Postgres>,
+    {
+        let base_query = format!(
+            r#"
+UPDATE
+    "{0}"
+SET
+    finished_at = NOW(),
+    status = 'completed'::job_status
+WHERE
+    "{0}".id = $2
+    AND queue = $1
+RETURNING
+    "{0}".*
+            "#,
+            queue_table
+        );
+
+        sqlx::query(&base_query)
+            .bind(&self.queue)
+            .bind(&self.id)
+            .execute(executor)
+            .await?;
+
+        Ok(CompletedJob {
             id: self.id,
             queue: self.queue,
-        }
+        })
     }
 
     /// Consume Job to fail it.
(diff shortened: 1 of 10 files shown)


hook-consumer/src/config.rs
```diff
@@ -72,4 +72,7 @@ pub struct RetryPolicyConfig {
 
     #[envconfig(default = "100000")]
     pub maximum_interval: EnvMsDuration,
+
+    #[envconfig(default = "default")]
+    pub retry_queue_name: String,
```
tomasfarias, line 77: Unsure if we should configure retries to a separate queue as a default. Separate queues require a different deployment pattern (two sets of consumers), so having to set this means the user acknowledges that such a deployment pattern is in use.

hook-common/src/retry.rs
```diff
@@ -1,55 +1,225 @@
...
+/// A retry policy to determine retry parameters for a job.
 pub struct RetryPolicy {
     /// Coefficient to multiply initial_interval with for every past attempt.
     pub backoff_coefficient: u32,
     /// The backoff interval for the first retry.
     pub initial_interval: time::Duration,
     /// The maximum possible backoff between retries.
     pub maximum_interval: Option<time::Duration>,
+    /// An optional queue to send WebhookJob retries to.
+    pub queue: Option<String>,
```
tomasfarias, line 16: Unfortunately, this means `RetryPolicy` is not `Copy` anymore.

tomasfarias: We get around it by using an `Arc` to wrap it.

hook-common/src/retry.rs
```diff
@@ -1,55 +1,225 @@
...
-                max_interval,
-            ),
+            (Some(duration), Some(max_interval)) => {
+                let min_interval_allowed = std::cmp::min(candidate_interval, max_interval);
+
+                if min_interval_allowed <= duration && duration <= max_interval {
+                    duration
+                } else {
+                    min_interval_allowed
+                }
```
tomasfarias, line 43: Expanding this makes it easier to follow.

hook-common/src/pgqueue.rs
```diff
@@ -441,42 +385,127 @@ RETURNING
...
+        S: serde::Serialize + std::marker::Sync + std::marker::Send,
+        E: sqlx::Executor<'c, Database = sqlx::Postgres>,
+    {
+        let json_error = sqlx::types::Json(error);
+        let base_query = format!(
+            r#"
+UPDATE
+    "{0}"
+SET
+    finished_at = NOW(),
```
bretthoerner, line 444: Should this be `attempted_at`? I feel like we discussed some confusion around these before, so I guess they're still confusing to me. :P

hook-consumer/src/config.rs
bretthoerner: > two sets of consumers

I wonder if we should make it easier on ourselves and make consumers handle multiple queues (I'm assuming it could be part of the dequeue `ORDER BY`)? Maybe that would make observability more annoying, though. Just thinking out loud.

bretthoerner (approved):
> Fix some bugs in retry.rs.

I was looking while reviewing but it didn't stand out to me, was there anything big?

hook-common/src/pgqueue.rs
bretthoerner: Ahhh, https://github.com/PostHog/rusty-hook/pull/7#discussion_r1426858428

It's more like `last_attempt_finished_at`, I had to refresh my memory.

hook-consumer/src/config.rs
tomasfarias: I wonder that too... Maybe we should invest some time on writing a benchmarking binary that we can run and test ideas on.

tomasfarias: At least to get some performance numbers. "Benchmarking observability" sounds more abstract...

tomasfarias:
> > Fix some bugs in retry.rs.
>
> I was looking while reviewing but it didn't stand out to me, was there anything big?

Some off-by-one retry interval computations. Behavior is now covered by tests which was missing.

hook-consumer/src/config.rs
tomasfarias: I'll go ahead and merge for now, we can revisit these ideas in the future as they could have potential.

PR #23, opened 2024-01-04
Change dequeue ORDER BY and rename finished_at

bretthoerner:
I guess these 2 are unrelated, but they were pretty small.

1. `last_attempt_finished_at` is a mouthful, so I'm open to better ideas, but I think it will help me/us when we come back to this code or are debugging things.
2. Ordering by `attempt` and then `scheduled_at` seems like a good short term upgrade until we decide whether stuff in https://github.com/PostHog/hog-rs/issues/2 is important yet.


hook-common/src/pgqueue.rs
```diff
@@ -135,7 +135,7 @@ impl<J, M> Job<J, M> {
 UPDATE
     "{0}"
 SET
-    finished_at = NOW(),
+    last_attempt_finished_at = NOW(),
```
tomasfarias, line 138: I don't mind longer names that have clearer meanings :+1:

migrations/20231129172339_job_queue_table.sql
```diff
@@ -9,21 +9,21 @@ CREATE TABLE job_queue(
...
     target TEXT NOT NULL
 );
 
 -- Needed for `dequeue` queries
-CREATE INDEX idx_queue_scheduled_at ON job_queue(queue, status, scheduled_at);
+CREATE INDEX idx_queue_scheduled_at ON job_queue(queue, status, scheduled_at, attempt);
 
 -- Needed for UPDATE-ing incomplete jobs with a specific target (i.e. slow destinations)
-CREATE INDEX idx_queue_target ON job_queue(queue, status, target);
+CREATE INDEX idx_queue_target ON job_queue(queue, status, target);
```
tomasfarias: Just going to add a newline here if you don't mind...
```suggestion
CREATE INDEX idx_queue_target ON job_queue(queue, status, target);

```

hook-janitor/src/webhooks.rs
```diff
@@ -198,10 +198,10 @@ impl WebhookCleaner {
...
-                (metadata->>'team_id')::bigint AS team_id,
-                (metadata->>'plugin_config_id')::bigint AS plugin_config_id,
-                count(*) as successes
+            SELECT DATE_TRUNC('hour', last_attempt_finished_at) AS hour,
+                   (metadata->>'team_id')::bigint AS team_id,
+                   (metadata->>'plugin_config_id')::bigint AS plugin_config_id,
+                   count(*) as successes
             FROM {0}
             WHERE status = 'completed'
               AND queue = $1
```
tomasfarias, line 207: Just consistent indenting. Not a huge fan of having sql in code, would prefer separate files, but that's another topic.
```suggestion
                AND queue = $1
```

tomasfarias: Just for consistent indenting? Seems like it's 4 spaces everywhere, but we added an extra 3 here:
```suggestion
                (metadata->>'team_id')::bigint AS team_id,
                (metadata->>'plugin_config_id')::bigint AS plugin_config_id,
                count(*) as successes
            FROM {0}
            WHERE status = 'completed'
                AND queue = $1
```

Indenting inline sql is rough...

tomasfarias (approved):
lgtm, just a minor suggestion regarding indents and spaces

PR #24, opened 2024-01-05
refactor: Deploy the migration container too

tomasfarias:
So that it can run migrations in production, not only in dev.

### DIFF
--- a/Dockerfile.migrate
+++ b/Dockerfile.migrate
@@ -1,7 +1,7 @@
 FROM docker.io/library/rust:1.74.0-buster as builder
 
 RUN apt update && apt install build-essential cmake -y
-RUN cargo install sqlx-cli --no-default-features --features native-tls,postgres --root /app/target/release/
+RUN cargo install sqlx-cli@0.7.3 --no-default-features --features native-tls,postgres --root /app/target/release/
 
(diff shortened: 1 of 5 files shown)


Dockerfile.migrate
```diff
@@ -0,0 +1,16 @@
+FROM docker.io/library/rust:1.74.0-buster as builder
+
+RUN apt update && apt install build-essential cmake -y
+RUN cargo install sqlx-cli --no-default-features --features native-tls,postgres --root /app/target/release/
```
xvello: We definitely want to pin the version here, as a bug in a new release could nuke a production database. Hopefully that syntax works?
```suggestion
RUN cargo install sqlx-cli@v0.7.3  --no-default-features --features native-tls,postgres --root /app/target/release/
```

tomasfarias: Yeah, good point. 

Tested locally and it seems like the syntax is without 'v':

```
cargo install sqlx-cli@v0.7.2                                                                                  
error: invalid value 'sqlx-cli@v0.7.2' for '[CRATE[@<VER>]]...': unexpected character 'v' while parsing major version number
```

xvello (approved):

PR #26, opened 2024-01-08
Minor query syntax/ordering cleanup

bretthoerner:
Minor tweaks to queries for readability (very subjective...).


tomasfarias (approved):
subjective, but I don't feel subjectively either way, :shipit:

PR #27, opened 2024-01-08
Add metrics, fix metrics endpoint, bump deps

bretthoerner:
I guess metrics macros slightly changed: https://crates.io/crates/metrics-macros

We don't _need_ to update, but I did it as part of debugging why metrics weren't working for me. Turns out that it was just the hardcoded `render()` call though: https://github.com/PostHog/rusty-hook/pull/27/files#diff-f71282eca22a13a695705e6e5b4a31b7497aefa59aba23dedb613b5691f62995L24

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,9 +19,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.8.6"
+version = "0.8.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


tomasfarias (approved):
I think I kind of prefer the new syntax with the `increment` method. Thank you.

bretthoerner:
> I think I kind of prefer the new syntax with the `increment` method. Thank you.

Yeah, the `count!("foo")` actually returns a handle you can reuse. It seems kind of doubtful the perf matters to us, but we could create and reuse existing metrics objects in the producer/consumer if we wanted to. It didn't seem worth the refactor right now.

tomasfarias:
Something to refactor later once our metrics get more complex, I agree!

PR #28, opened 2024-01-08
Get total row counts in janitor for metrics and consistency checking

bretthoerner:
I split this one off since it's a bit separate, but it's pointing at (and depends on) https://github.com/PostHog/rusty-hook/pull/27

I feel like these count queries should be very cheap, and they serve two purposes:

1. Metrics/logging of actual row count, rather than the `GROUP BY` (`agg`) row counts we already do.
2. A simple way to check that the count of rows we're deleting match the count we observed when we started doing the work.

### DIFF
--- a/hook-janitor/src/webhooks.rs
+++ b/hook-janitor/src/webhooks.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use async_trait::async_trait;
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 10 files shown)


hook-janitor/src/webhooks.rs
```diff
@@ -315,33 +346,51 @@ impl WebhookCleaner {
...
 
-        let mut rows_processed = 0;
+        let mut rows_deleted = 0;
         if completed_agg_row_count + failed_agg_row_count != 0 {
-            rows_processed = self.delete_observed_rows(&mut tx).await?;
+            rows_deleted = self.delete_observed_rows(&mut tx).await?;
+
+            if rows_deleted != completed_row_count + failed_row_count {
+                // This should never happen, but if it does, we want to know about it (and abort the
+                // txn).
```
tomasfarias, line 385: I'm somewhat afraid that we are not aborting the transaction here. Since we take a `&mut` instead of taking ownership of `tx` the caller could still commit the transaction even on error, once we drop the reference.

Could we take ownership of `tx` and call `rollback` here? Similar to `commit_txn` below.

tomasfarias: I agree that this should never happen, but in the event it does, I would like to ensure we clean up any loose ends properly :sparkles:

bretthoerner: Good idea, I shouldn't depend on `Drop` for something so dangerous.

tomasfarias: Thank you!

tomasfarias (approved):

PR #29, opened 2024-01-09
refactor: Rename producer and consumer images

tomasfarias:
Rename images to match their k8s service counterparts.


bretthoerner (approved):
I wonder if we should rename things internally too, to avoid confusion.

tomasfarias:
@bretthoerner Yeah, we should, just more complicated as we have plastered the names all over. Was hoping to tackle that in a follow-up PR.

tomasfarias:
The follow-up PR: #31

PR #31, opened 2024-01-09
refactor: Rename producer for api and consumer for worker

tomasfarias:
To align with service names and image names (see #29), here we rename every instance of producer to api and every instance of consumer to worker. Had to manually verify a lot as, for example, `rdkafka::producer` should still be producer.

~Tests may be failing and I'll fix them.~
EDIT: They should now be okay...


hook-janitor/src/config.rs
```diff
@@ -30,10 +30,10 @@ pub struct Config {
 #[derive(Envconfig, Clone)]
 pub struct KafkaConfig {
     #[envconfig(default = "20")]
-    pub kafka_producer_linger_ms: u32, // Maximum time between producer batches during low traffic
+    pub kafka_producer_linger_ms: u32, // Maximum time between api batches during low traffic
 
     #[envconfig(default = "400")]
-    pub kafka_producer_queue_mib: u32, // Size of the in-memory producer queue in mebibytes
+    pub kafka_producer_queue_mib: u32, // Size of the in-memory api queue in mebibytes
```
tomasfarias: These two shouldn't be renamed...

hook-janitor/src/webhooks.rs
```diff
@@ -511,26 +511,22 @@ mod tests {
...
-        let consumer: StreamConsumer = ClientConfig::new()
+        let worker: StreamConsumer = ClientConfig::new()
             .set("bootstrap.servers", mock_cluster.bootstrap_servers())
             .set("group.id", "mock")
             .set("auto.offset.reset", "earliest")
             .create()
-            .expect("failed to create mock consumer");
-        consumer.subscribe(&[APP_METRICS_TOPIC]).unwrap();
+            .expect("failed to create mock worker");
+        worker.subscribe(&[APP_METRICS_TOPIC]).unwrap();
```
tomasfarias: Funnily enough, this works, but also shouldn't be renamed...

bretthoerner (approved):
A few things caught my eye, it was easier to just push: https://github.com/PostHog/rusty-hook/pull/31/commits/e3084321c28ef06e13a52168b4d3171f5d011400

xvello (approved):

PR #32, opened 2024-01-09
Login to Docker before QEMU

bretthoerner:
I missed that the QEMU setup is doing Docker pulls.

### DIFF
--- a/.github/workflows/docker-hook-api.yml
+++ b/.github/workflows/docker-hook-api.yml
@@ -17,6 +17,19 @@ jobs:
 
       - name: Check Out Repo
         uses: actions/checkout@v3
+      
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-hook-api.yml
```diff
@@ -17,6 +17,19 @@ jobs:
 
       - name: Check Out Repo
         uses: actions/checkout@v3
+      
+      - name: Login to DockerHub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
tomasfarias, line 25: I don't seem to have repo access to verify these secrets are loaded in. Do you?

tomasfarias: I guess they are if we were doing this below!

tomasfarias (approved):

.github/workflows/docker-hook-api.yml
bretthoerner: Yeah sorry, this was discussed in Slack, they are organization level so we are all set. :)

PR #33, opened 2024-01-10
Sync default ports with charts

bretthoerner:
Charts uses `BIND_PORT` but we may as well avoid the collisions locally.

https://github.com/PostHog/charts/pull/642/files


tomasfarias (approved):
Any chance of collision if we run multiple workers? Anyways, I don't think that would be solved with default values.

bretthoerner:
> Any chance of collision if we run multiple workers? Anyways, I don't think that would be solved with default values.

Yeah, they would definitely collide, but if you want to do that you'll need to override the port on some of them.

xvello (approved):

PR #39, opened 2024-01-10
feat: Set idle tx timeout in migration

tomasfarias:
This [setting](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-IDLE-IN-TRANSACTION-SESSION-TIMEOUT) is disabled by default. We rely on it to clean up things on transactional mode, so let's set it in a migration.

### DIFF
--- a/migrations/20240110180056_set_idle_in_transaction_timeout.sql
+++ b/migrations/20240110180056_set_idle_in_transaction_timeout.sql
@@ -1,2 +1,2 @@
 -- If running worker in transactional mode, this ensures we clean up any open transactions.
-SET idle_in_transaction_session_timeout='2min';
+ALTER USER current_user SET idle_in_transaction_session_timeout = '2min';
--- a/migrations/20240110180056_set_idle_in_transaction_timeout.sql
+++ b/migrations/20240110180056_set_idle_in_transaction_timeout.sql
@@ -1,0 +1,2 @@
+-- If running worker in transactional mode, this ensures we clean up any open transactions.
+SET idle_in_transaction_session_timeout='2min';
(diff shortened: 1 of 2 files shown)


migrations/20240110180056_set_idle_in_transaction_timeout.sql
```diff
@@ -0,0 +1,2 @@
+-- If running worker in transactional mode, this ensures we clean up any open transactions.
+SET idle_in_transaction_session_timeout='2min';
```
bretthoerner: I think sqlx sets one?

https://docs.rs/sqlx-core/0.7.3/src/sqlx_core/pool/options.rs.html#147

We could override it when we make `PoolOptions` if we want to change it.

---

But if we want to change it on the cluster, I'm struggling to find proof that this will do it. I thought this was a connection level setting? Does putting this in a migration really change the DB/cluster default?

tomasfarias: I saw that one, but that `idle_timeout` refers to a connection idle timeout enforced by (I believe) sqlx, at least that's what I gather looking at the docs: https://docs.rs/sqlx/latest/sqlx/pool/struct.PoolOptions.html#method.idle_timeout

Although valuable to set that one too, the reason to have a tx timeout enforced by PG is to have PG clean up any transactions left open after a pod crashes or becomes unresponsive for whatever other reason.

> But if we want to change it on the cluster, I'm struggling to find proof that this will do it. I thought this was a connection level setting? Does putting this in a migration really change the DB/cluster default?

I can update the query to make it affect the current user. As long as the same user is used to run migrations and the services, we'll be fine. Ideally this should be updated in the postgres configuration though.

bretthoerner (approved):

PR #41, opened 2024-01-10
Drop target from Prom labels

bretthoerner:
[As discussed in Slack](https://posthog.slack.com/archives/C066YCDAWG2/p1704832861479619), this will be too high cardinality in practice for Prometheus.

I'm investigating how to have useful metrics in app_metrics (or somewhere).


xvello (approved):
agreed. should be added to traces when we implement tracing

PR #43, opened 2024-01-10
Fix app_metrics topic, don't encode null error_uuid

bretthoerner:
Messages in Kafka look good, but I guess was I putting them in the wrong topic.

### DIFF
--- a/hook-janitor/src/config.rs
+++ b/hook-janitor/src/config.rs
@@ -44,7 +44,7 @@ pub struct KafkaConfig {
     #[envconfig(default = "false")]
     pub kafka_tls: bool,
 
-    #[envconfig(default = "app_metrics")]
+    #[envconfig(default = "clickhouse_app_metrics")]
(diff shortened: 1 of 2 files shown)


hook-janitor/src/config.rs
```diff
@@ -44,7 +44,7 @@ pub struct KafkaConfig {
     #[envconfig(default = "false")]
     pub kafka_tls: bool,
 
-    #[envconfig(default = "app_metrics")]
+    #[envconfig(default = "clickhouse_app_metrics")]
```
xvello, line 47: Need to double-check we're not also setting it in the chart. let's remove from the chart if that's the case, we probably won't need to update

xvello (approved):

hook-janitor/src/config.rs
bretthoerner: Good catch, it's in the chart.

PR #44, opened 2024-01-11
Remove idle_in_transaction_session_timeout migration

bretthoerner:
So because this is ALTERing the user, it races and makes concurrent tests fail with `tuple concurrently updated`.

I tried adding an advisory lock to no avail.

This should change it in the DB: https://github.com/PostHog/posthog-cloud-infra/pull/2433


xvello (approved):

tomasfarias (approved):
works for me, thank you for fixing this!

PR #45, opened 2024-01-12
Use signed integers for plugin_id/plugin_config_id

bretthoerner:
The plan is to use sentinel values -1 and -2 (and maybe others in the future) for fake "plugins" for things like Webhooks that are sent by Action integrations, and not real plugins.

https://posthog.slack.com/archives/C066YCDAWG2/p1705010221524319?thread_ts=1705007769.632829&cid=C066YCDAWG2

### DIFF
--- a/hook-common/src/kafka_messages/app_metrics.rs
+++ b/hook-common/src/kafka_messages/app_metrics.rs
@@ -51,7 +51,7 @@ pub struct AppMetric {
     )]
     pub timestamp: DateTime<Utc>,
     pub team_id: u32,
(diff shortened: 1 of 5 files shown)


xvello (approved):

tomasfarias (approved):
I'd prefer if we had our own custom numerical type that can understand and enforce the sentinel values, but maybe for later.

bretthoerner:
> I'd prefer if we had our own custom numerical type that can understand and enforce the sentinel values, but maybe for later.

Good idea, I'll put it on my ideas list for now.

PR #46, opened 2024-01-15
feat: make sure metrics cover all axum endpoints

xvello:
`hooks-api` does not currently export metrics for the `/webhook` endpoint, as the layer is added to the router before this endpoint:

- Change `setup_metrics_router` into `setup_metrics_routes` that is called with the `Router` already setup with routes to report metrics on
- Change the three roles to use this changed signature

### DIFF
--- a/hook-common/src/metrics.rs
+++ b/hook-common/src/metrics.rs
@@ -16,11 +16,11 @@ pub async fn serve(router: Router, bind: &str) -> Result<(), std::io::Error> {
     Ok(())
 }
 
-/// Build a Router for a metrics endpoint.
-pub fn setup_metrics_router() -> Router {
+/// Add the prometheus endpoint and middleware to a router, should be called last.
+pub fn setup_metrics_routes(router: Router) -> Router {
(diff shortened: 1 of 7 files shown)


tomasfarias (approved):
lgtm :shipit:

hook-common/src/metrics.rs
```diff
@@ -16,11 +16,11 @@ pub async fn serve(router: Router, bind: &str) -> Result<(), std::io::Error> {
     Ok(())
 }
 
-/// Build a Router for a metrics endpoint.
-pub fn setup_metrics_router() -> Router {
+/// Add the prometheus endpoint and middleware to a router, should be called last.
```
tomasfarias, line 19: I'm sure there is a build pattern thing we can code up here to enforce this is called last, but I have a TODO to revisit this later.

Taking a `Router` is a good start, thanks!

xvello: Agreed, this should be redone better later, but it's not the highest priority right now.

PR #47, opened 2024-01-16
feat: add readiness and liveness endpoints to all roles

xvello:
In my experience, it’s better not to fail liveness or readiness on missing dependencies, because restarting pods is less information than proper http errors / error logs. This PR mirrors capture’s behaviour:

- having all readiness probes to a simple index endpoint: axum is started last, and the pod should be ready to accept requests as soon as axum is ready
- monitor the long-lived async loops (janitor loop, worker loop, rdkafka producer) on the liveness: if they stop / hang for too long, the process will be restarted. For api, there’s currently no background loop, so liveness = index for now

For simplicity, I'm copying `health.rs` over from capture-rs, we should:
  - rework its api to not depend on the `time` crate
  - share the crate between the codebases

## Tests:

Janitor:
```
$ curl localhost:3302/_readiness
rusty-hook janitor
$ curl localhost:3302/_liveness
HealthStatus { healthy: true, components: {"cleanup_loop": HealthyUntil(2024-01-16 16:07:54.088147 +00:00:00), "rdkafka": HealthyUntil(2024-01-16 16:07:24.087438 +00:00:00)} }
```

Worker:
```
$ curl localhost:3301/_readiness
rusty-hook worker
$ curl localhost:3301/_liveness
HealthStatus { healthy: true, components: {"worker": HealthyUntil(2024-01-16 16:09:22.245115 +00:00:00)} }
```

API: 
```
$ curl localhost:3300/_readiness
rusty-hook api
$ curl localhost:3300/_liveness
rusty-hook api
```

### DIFF
--- a/hook-worker/src/worker.rs
+++ b/hook-worker/src/worker.rs
@@ -2,6 +2,7 @@ use std::collections;
 use std::sync::Arc;
 use std::time;
 
+use hook_common::health::HealthHandle;
 use hook_common::{
     pgqueue::{Job, PgJob, PgJobError, PgQueue, PgQueueError, PgQueueJob, PgTransactionJob},
     retry::RetryPolicy,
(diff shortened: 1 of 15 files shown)


hook-worker/src/worker.rs
```diff
@@ -140,11 +141,16 @@ impl<'p> WebhookWorker<'p> {
...
+    pub async fn run(
+        &self,
+        transactional: bool,
+        liveness: HealthHandle,
+    ) -> Result<(), WorkerError> {
         let semaphore = Arc::new(sync::Semaphore::new(self.max_concurrent_jobs));
 
         if transactional {
             loop {
+                liveness.report_healthy().await;
```
bretthoerner: These might need to move into `wait_for_job`. Those loop internally and don't yield a job until one is found.

I suppose it's unlikely to take that long in prod, but it could happen, and it'll probably happen in `dev`.

bretthoerner (approved):

hook-worker/src/worker.rs
xvello: Good catch, thanks! Fixed it

PR #49, opened 2024-01-16
Add queue depth gauges

bretthoerner:


hook-janitor/src/webhooks.rs
```diff
@@ -354,6 +389,12 @@ impl WebhookCleaner {
         // of rows in memory. It seems unlikely we'll need to paginate, but that can be added in the
         // future if necessary.
 
+        let queue_depth = self.get_queue_depth().await?;
+        metrics::gauge!("queue_depth_oldest_created_at")
```
xvello: I think there's value in computing it for both untried and retried, as we're most likely more interested in untried (happy path)

hook-janitor/src/webhooks.rs
```diff
@@ -354,6 +389,12 @@ impl WebhookCleaner {
...
+            .set(queue_depth.oldest_created_at.timestamp() as f64);
+        metrics::gauge!("queue_depth_untried").set(queue_depth.count_untried as f64);
+        metrics::gauge!("queue_depth_retries").set(queue_depth.count_retries as f64);
```
xvello: nit: could be a single metric name with a tag, for easier sums

xvello (approved):

PR #50, opened 2024-01-16
feat: add webhook_worker_saturation_percent metric for autoscaling

xvello:
We'll want to implement autoscaling of the worker deployment via Keda. Keda works best when it autoscales on a metric that truly measures the pods' bottleneck.

For the delivery workers, our bottleneck is the semaphore in the worker loop, that holds `max_concurrent_jobs` permits. This PR exports a `webhook_worker_saturation_percent` gauge, that is 0 when no in-flight tasks are running, and 1 when we hit `webhook_worker_saturation_percent` and will stop de-queuing payloads from PG. Having a percentage will allow us to easily setup a Keda scaler to target an avg 80% saturation of the worker pool.

The placement of `report_semaphore_utilization()` is not perfect (we should ideally update it when we're dropping permits too), but I feel like this should be a close-enough approximation for the purposes of autoscaling.

### DIFF
--- a/hook-worker/src/worker.rs
+++ b/hook-worker/src/worker.rs
@@ -149,9 +149,14 @@ impl<'p> WebhookWorker<'p> {
     /// Run this worker to continuously process any jobs that become available.
     pub async fn run(&self, transactional: bool) -> Result<(), WorkerError> {
         let semaphore = Arc::new(sync::Semaphore::new(self.max_concurrent_jobs));
+        let report_semaphore_utilization = || {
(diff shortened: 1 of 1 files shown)


bretthoerner (approved):
> The placement of report_semaphore_utilization() is not perfect (we should ideally update it when we're dropping permits too), but I feel like this should be a close-enough approximation for the purposes of autoscaling.

Yeah, I was a little nervous about reporting it before we even acquire the semaphore, but I guess under active load this will report a close enough approximation (by observing the count globally) anyway.

PR #51, opened 2024-01-16
Log error rather than exiting process on dequeue error

bretthoerner:
@xvello and I were talking, and it seems better to log an error and retry if PG has a hiccup. Crashing and restarting the pod is pretty expensive, and monitoring should catch if we aren't making progress on the queue.

### DIFF
--- a/hook-worker/src/worker.rs
+++ b/hook-worker/src/worker.rs
@@ -115,17 +115,20 @@ impl<'p> WebhookWorker<'p> {
     }
 
     /// Wait until a job becomes available in our queue.
-    async fn wait_for_job<'a>(
-        &self,
-    ) -> Result<PgJob<WebhookJobParameters, WebhookJobMetadata>, WorkerError> {
+    async fn wait_for_job<'a>(&self) -> PgJob<WebhookJobParameters, WebhookJobMetadata> {
         let mut interval = tokio::time::interval(self.poll_interval);
 
         loop {
             interval.tick().await;
             self.liveness.report_healthy().await;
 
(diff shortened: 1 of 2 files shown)


hook-worker/src/worker.rs
```diff
@@ -115,39 +115,47 @@ impl<'p> WebhookWorker<'p> {
...
             self.liveness.report_healthy().await;
 
-            if let Some(job) = self.queue.dequeue(&self.name).await? {
-                return Ok(job);
+            match self.queue.dequeue(&self.name).await {
+                Ok(Some(job)) => return job,
+                Ok(None) => continue,
+                Err(error) => {
+                    error!("error while trying to dequeue job: {}", error);
+                    continue;
```
tomasfarias, line 130: Maybe we could have an extra `interval.tick` or similar here to slow down if we encounter an error... But maybe for later

tomasfarias (approved):
lgtm, maybe we can have a fancier policy for dealing with errors (i.e. incremental backoff, timeout, etc), but this works fine. Thank you!

hook-worker/src/worker.rs
bretthoerner: We could also just do a naive sleep in this branch. We're polling at 100ms currently and I could just sleep for a second to add some backoff?

bretthoerner: Eh, I'll see how it goes. For now, we have the 100ms (as currently configured) interval, which prevents us from being in a super hot retry loop.

PR #52, opened 2024-01-16
Remove queue filters from janitor

bretthoerner:
I shouldn't have added the `queue` here to begin with. The janitor should process all rows in the table, so that we don't have to think about and configure the janitor in line with the main queue and the retry queue.

If for some reason we want to partition work so the janitor doesn't touch certain rows, we should use a different shard (or even just table).

---

The alternative to this is to change the `QUEUE_NAME` envconfig to `QUEUE_NAMES` (note the `S`) list and filter on multiple queues. That was my initial plan (to support [the retry queue](https://github.com/PostHog/charts/blob/427928bac53f4e4a64e9bfa7e0e9c4e7141ebd81/charts/hooks/values.yaml#L77)) but it seems entirely pointless to even have to filter.

### DIFF
--- a/hook-janitor/src/config.rs
+++ b/hook-janitor/src/config.rs
@@ -11,9 +11,6 @@ pub struct Config {
     #[envconfig(default = "postgres://posthog:posthog@localhost:15432/test_database")]
     pub database_url: String,
 
-    #[envconfig(default = "default")]
-    pub queue_name: String,
-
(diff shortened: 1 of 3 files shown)


tomasfarias (approved):

tomasfarias:
We can always re-add this later. As it stands, I can't think of reason to have janitor scoped only to one queue (or queues).

PR #53, opened 2024-01-16
Make retry_queue_name truly optional from env to main

bretthoerner:
This was setting our retry queue to `default` when we don't set it, which would have been confusing.

### DIFF
--- a/hook-worker/src/main.rs
+++ b/hook-worker/src/main.rs
@@ -23,13 +23,13 @@ async fn main() -> Result<(), WorkerError> {
         .register("worker".to_string(), time::Duration::seconds(60)) // TODO: compute the value from worker params
         .await;
 
-    let retry_policy_builder = RetryPolicy::build(
+    let mut retry_policy_builder = RetryPolicy::build(
         config.retry_policy.backoff_coefficient,
         config.retry_policy.initial_interval.0,
     )
     .maximum_interval(config.retry_policy.maximum_interval.0);
 
-    let retry_policy_builder = match &config.retry_policy.retry_queue_name {
+    retry_policy_builder = match &config.retry_policy.retry_queue_name {
         Some(retry_queue_name) if !retry_queue_name.is_empty() => {
             retry_policy_builder.queue(retry_queue_name)
         }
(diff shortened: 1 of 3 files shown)


hook-worker/src/main.rs
```diff
@@ -23,13 +23,19 @@ async fn main() -> Result<(), WorkerError> {
...
         config.retry_policy.backoff_coefficient,
         config.retry_policy.initial_interval.0,
     )
-    .maximum_interval(config.retry_policy.maximum_interval.0)
-    .queue(&config.retry_policy.retry_queue_name)
-    .provide();
+    .maximum_interval(config.retry_policy.maximum_interval.0);
+
+    retry_policy_builder = match &config.retry_policy.retry_queue_name {
+        Some(retry_queue_name) if !retry_queue_name.is_empty() => {
```
tomasfarias: Seems like we are also making it so `RETRY_QUEUE_NAME` cannot be empty. Nothing against this necessarily, but `QUEUE_NAME` can be empty string, so there is a bit of a disconnect there.

Personally, I'd have a wrapper type in config, `NonEmptyString` or something, and use it in both, similar to `EnvMsDuration`.

tomasfarias (approved):

hook-worker/src/main.rs
bretthoerner: Good point.

It's actually an issue I have with `envconfig` in general. It seems very easy to set `FOO=` in the env and think you are making it "unset" and you're going to get the default.

We set (many of) the env vars in charts [regardless of whether the config has a value](https://github.com/PostHog/charts/blob/595939842f130ba04e5c73cc8cec059f09233144/charts/hooks/templates/_helpers.tpl#L88-L89). So if you drop the config, we're actually setting an empty string in the env. Maybe I should just fix it there with lots of `if`s in the chart?

Or I could go the `NonEmptyString` route everywhere. Or maybe I'm overthinking it.

xvello: Yup, I'm pretty unhappy with `envconfig` too, notably because it makes capture integration tests clunky (cannot set a few overrides and fill with defaults). I went with a pretty simple lib at first, but happy to switch to another, as long as we keep the setup simple and fool-proof

tomasfarias: E.g.: https://github.com/PostHog/rusty-hook/pull/55

bretthoerner: Oh nice, I had just removed the non-empty string check so I don't conflate two issues here.

I'll merge this and check out your change against main.

PR #54, opened 2024-01-17
fix: fix janitor get_queue_depth when queue is empty

xvello:
Janitor currently fails on an empty queue with:

> 2024-01-17T10:33:38.342667Z ERROR hook_janitor::webhooks: WebhookCleaner::cleanup failed error=GetQueueDepthError { error: ColumnDecode { index: "\"count_untried\"", source: UnexpectedNullError } }

Coalesce a zero to avoid returning a null to rust

### DIFF
--- a/hook-janitor/src/webhooks.rs
+++ b/hook-janitor/src/webhooks.rs
@@ -513,6 +513,7 @@ mod tests {
     use rdkafka::consumer::{Consumer, StreamConsumer};
     use rdkafka::mocking::MockCluster;
     use rdkafka::producer::{DefaultProducerContext, FutureProducer};
+    use rdkafka::types::{RDKafkaApiKey, RDKafkaRespErr};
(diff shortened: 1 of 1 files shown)


tomasfarias (approved):
Should be good, I would just ask that we copy the `test_cleanup_impl` test below, but remove all the test data generation code and assert no metrics are produced. Just to cover this case, although I think it should now work.

hook-janitor/src/webhooks.rs
```diff
@@ -754,6 +755,47 @@ mod tests {
         check_app_metric_vector_equality(&expected_app_metrics, &received_app_metrics);
     }
 
+    #[sqlx::test(migrations = "../migrations")]
+    async fn test_cleanup_impl_empty_queue(db: PgPool) {
```
tomasfarias, line 759: Thank you!

PR #55, opened 2024-01-17
 refactor: Use new NonEmptyString config type  for queue names

tomasfarias:
Looks a bit cleaner and now `RETRY_QUEUE_NAME` and `QUEUE_NAME` are consistent, but not super necessary.

### DIFF
--- a/hook-worker/src/config.rs
+++ b/hook-worker/src/config.rs
@@ -18,7 +18,7 @@ pub struct Config {
     pub worker_name: String,
 
     #[envconfig(default = "default")]
-    pub queue_name: String,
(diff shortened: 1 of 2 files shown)


bretthoerner:
Ah interesting, in my head I was thinking `FOO=` would just become a `None` rather than an `Err`, but I think being loud about this is probably better.

bretthoerner (approved):
Makes sense, I think we'll (eventually) want this in `common` to we can use it elsewhere, but we can decide on things like that when we merge capture/rusty-hook, because maybe we'll pick a different config system at the offsite? 🤷‍♂️

PR #56, opened 2024-01-18
Dequeue multiple items at a time (to a configured limit)

bretthoerner:
The dequeue query is especially expensive if we're just plucking out one row.

For the non-transactional implementation this is easy, just return more than 1, and acquire connections as needed to update to the DB after jobs run.

For the transactional implementation, we need to share the txn and only commit when all jobs are done (even if they failed, that's fine). I tried a number of ways, and am going to take a break and get some 👀 before I look at this any more.

---

What I have works, but it's a little ugly. I've been struggling to find a clean way to test it (I'd like to break the txn in the middle and prove things work). The happy path works fine.

Why not `commit` in `impl Drop`?
1. You can't call async code in Drop, and commit is async.
2. We don't want to block the event loop in drop to get around that.
3. You can't move something with a non-static lifetime (the transaction) into an async block.

Maybe putting things on a channel would work, but it seemed like even more refactoring and services to manage.

A much larger refactor would be to expose the transaction to the worker. In a perfect world, we'd receive a `Vec` of jobs, run them all in parallel, and just commit at the end of a plain old normal async fn.

Curious for thoughts.

### DIFF
--- a/hook-common/src/pgqueue.rs
+++ b/hook-common/src/pgqueue.rs
@@ -30,6 +30,8 @@ pub enum PgQueueError {
 pub enum PgJobError<T> {
     #[error("retry is an invalid state for this PgJob: {error}")]
     RetryInvalidError { job: T, error: String },
+    #[error("connection failed with: {error}")]
+    ConnectionError { error: sqlx::Error },
     #[error("{command} query failed with: {error}")]
     QueryError { command: String, error: sqlx::Error },
     #[error("transaction {command} failed with: {error}")]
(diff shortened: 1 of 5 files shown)


bretthoerner:
> A much larger refactor would be to expose the transaction to the worker. In a perfect world, we'd receive a Vec of jobs, run them all in parallel, and just commit at the end of a plain old normal async fn.

Actually, maybe this would be less refactor than I'm thinking. Rather than returning a `Vec<Job>` we could return something like a `Batch`, which contains a `Vec<Job>`, each of which could be completed/failed/retried, and then the Batch itself could carry a `commit` method to call at the end. And this whole object could be processed by a `process_webhook_jobs` (note the `s`) function similar to our current `process_webhook_job`.

Thanks for attending my rubberduck session.

PR #58, opened 2024-01-18
Change created_at to scheduled_at for metrics

bretthoerner:


hook-janitor/src/webhooks.rs
```diff
@@ -374,12 +374,12 @@ impl WebhookCleaner {
         // future if necessary.
 
         let queue_depth = self.get_queue_depth().await?;
-        metrics::gauge!("queue_depth_oldest_created_at_untried")
-            .set(queue_depth.oldest_created_at_untried.timestamp() as f64);
+        metrics::gauge!("queue_depth_oldest_scheduled_at_untried")
```
xvello: nit: we could reuse `queue_depth`'s `status` tag values for consistency, instead of creating two timeseries

xvello (approved):

PR #60, opened 2024-01-31
Dequeue multiple items at a time

bretthoerner:
This amortizes the work done in both querying PG, and in the number of Tokio tasks spawned (we spawn 1 per *batch* rather than 1 per *job*).

The gist is that `PgJob` and `PgTransactionJob` are now wrapped in `PgBatch` and `PgTransactionBatch` respectively. Each contains a `Vec` of jobs. `dequeue` takes a `limit` parameter and returns up to that many jobs at once.

For non-txn jobs, everything is easy and mostly the same. I no longer hold a connection open for non-txn jobs (which is, to me, the benefit of non-txn jobs), so each job completes, then grabs a conn from the pool and does the query to finalize its work.

For txn jobs, we need to share the single txn object among them, and with the batch itself. Each one completes and grabs the txn via a mutex to do its work. After all work is done, the top-level Worker code is what calls commit.

In the future both of these could be optimized: the Worker code could do all Webhook stuff in parallel, and then updating the DB could be done in a batch query (or queries) rather than 1 per job. I'd rather do that as part of a future PR.

### DIFF
--- a/hook-common/src/pgqueue.rs
+++ b/hook-common/src/pgqueue.rs
@@ -1,14 +1,17 @@
 //! # PgQueue
 //!
 //! A job queue implementation backed by a PostgreSQL table.
-use std::str::FromStr;
+use std::ops::DerefMut;
 use std::time;
+use std::{str::FromStr, sync::Arc};
 
 use async_trait::async_trait;
 use chrono;
 use serde;
 use sqlx::postgres::{PgConnectOptions, PgPool, PgPoolOptions};
 use thiserror::Error;
+use tokio::sync::Mutex;
+use tracing::error;
 
 /// Enumeration of errors for operations with PgQueue.
 /// Errors that can originate from sqlx and are wrapped by us to provide additional context.
@@ -24,6 +27,8 @@ pub enum PgQueueError {
     ParseJobStatusError(String),
     #[error("{0} is not a valid HttpMethod")]
     ParseHttpMethodError(String),
+    #[error("transaction was already closed")]
+    TransactionAlreadyClosedError,
 }
 
 #[derive(Error, Debug)]
@@ -36,6 +41,8 @@ pub enum PgJobError<T> {
     QueryError { command: String, error: sqlx::Error },
     #[error("transaction {command} failed with: {error}")]
     TransactionError { command: String, error: sqlx::Error },
+    #[error("transaction was already closed")]
+    TransactionAlreadyClosedError,
 }
 
 /// Enumeration of possible statuses for a Job.
@@ -222,6 +229,11 @@ pub struct PgJob<J, M> {
     pub pool: PgPool,
 }
 
+// Container struct for a batch of PgJobs.
+pub struct PgBatch<J, M> {
+    pub jobs: Vec<PgJob<J, M>>,
+}
+
 impl<J: std::marker::Send, M: std::marker::Send> PgJob<J, M> {
     async fn acquire_conn(
         &mut self,
@@ -304,7 +316,72 @@ impl<J: std::marker::Send, M: std::marker::Send> PgQueueJob for PgJob<J, M> {
 #[derive(Debug)]
 pub struct PgTransactionJob<'c, J, M> {
     pub job: Job<J, M>,
-    pub transaction: sqlx::Transaction<'c, sqlx::postgres::Postgres>,
+
+    /// The open transaction this job came from. If multiple jobs were queried at once, then this
+    /// transaction will be shared between them (across async tasks and threads as necessary). See
+    /// below for more information.
+    shared_txn: Arc<Mutex<SharedTxn<'c>>>,
+}
+
+// Container struct for a batch of PgTransactionJob. Includes a reference to the shared transaction
+// for committing the work when all of the jobs are finished.
+pub struct PgTransactionBatch<'c, J, M> {
+    pub jobs: Vec<PgTransactionJob<'c, J, M>>,
+
+    /// The open transaction the jobs in the Vec came from. This should be used to commit or
+    /// rollback when all of the work is finished.
+    shared_txn: Arc<Mutex<SharedTxn<'c>>>,
+}
+
+impl<'c, J, M> PgTransactionBatch<'_, J, M> {
+    pub async fn commit(&mut self) -> PgQueueResult<()> {
+        let mut txn_guard = self.shared_txn.lock().await;
+        txn_guard.commit().await
+    }
+}
+
+/// A shared transaction. Exists as a named type so that it can have some helper methods.
+#[derive(Debug)]
+struct SharedTxn<'c> {
+    /// The actual transaction object. If a transaction error occurs (e.g. the connection drops)
+    /// then the transaction should be dropped so that other jobs don't each try to do DB work that
+    /// is bound to fail.
+    raw_txn: Option<sqlx::Transaction<'c, sqlx::postgres::Postgres>>,
+}
+
+impl SharedTxn<'_> {
+    /// Commits and then drops the transaction.
+    pub async fn commit(&mut self) -> PgQueueResult<()> {
+        if self.raw_txn.is_none() {
+            return Err(PgQueueError::TransactionAlreadyClosedError);
+        }
+
+        let txn = self.raw_txn.take().unwrap();
+        txn.commit().await.map_err(|e| PgQueueError::QueryError {
+            command: "COMMIT".to_owned(),
+            error: e,
+        })?;
+
+        Ok(())
+    }
+
+    /// See `raw_txn` above, this should be called when a transaction error occurs (e.g. the
+    /// connection drops) so that other jobs don't each try to do DB work that is bound to fail.
+    fn drop_txn(&mut self) {
+        self.raw_txn = None;
+    }
+}
+
+impl SharedTxn<'_> {
+    /// Helper to get the transaction reference if it exists, for use in sqlx queries.
+    /// If it doesn't exist, that means a previous error occurred and the transaction has been
+    /// dropped.
+    fn get_txn_ref(&mut self) -> Option<&mut sqlx::PgConnection> {
+        match self.raw_txn.as_mut() {
+            Some(txn) => Some(txn.deref_mut()),
+            None => None,
+        }
+    }
 }
 
 #[async_trait]
@@ -312,22 +389,20 @@ impl<'c, J: std::marker::Send, M: std::marker::Send> PgQueueJob for PgTransactio
     async fn complete(
         mut self,
     ) -> Result<CompletedJob, PgJobError<Box<PgTransactionJob<'c, J, M>>>> {
-        let completed_job = self
-            .job
-            .complete(&mut *self.transaction)
-            .await
-            .map_err(|error| PgJobError::QueryError {
+        let mut txn_guard = self.shared_txn.lock().await;
+
+        let txn_ref = txn_guard
+            .get_txn_ref()
+            .ok_or(PgJobError::TransactionAlreadyClosedError)?;
+
+        let completed_job = self.job.complete(txn_ref).await.map_err(|error| {
+            txn_guard.drop_txn();
+
+            PgJobError::QueryError {
                 command: "UPDATE".to_owned(),
                 error,
-            })?;
-
-        self.transaction
-            .commit()
-            .await
-            .map_err(|error| PgJobError::TransactionError {
-                command: "COMMIT".to_owned(),
-                error,
-            })?;
+            }
+        })?;
 
         Ok(completed_job)
     }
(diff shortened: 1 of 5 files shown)


hook-worker/src/worker.rs
```diff
@@ -165,65 +172,87 @@ impl<'p> WebhookWorker<'p> {
         if transactional {
```
bretthoerner, line 182: Note that the duplicated code for txn vs non-txn predated this PR. Refactoring to not have any duplication would be a lot of work, I think, and maybe/hopefully we'll drop one of these modes soon anyway?

xvello: Yup, my understanding is that we want to keep our options open while we're evaluating the performance profile of the workload, but we should drop one of them before the end of Q1. One of them is basically dead code I would not feel comfortable switching to anyway.

hook-worker/src/config.rs
```diff
@@ -34,6 +34,9 @@ pub struct Config {
 
     #[envconfig(default = "true")]
     pub transactional: bool,
+
+    #[envconfig(default = "10")]
+    pub dequeue_count: u32,
```
xvello: nit: as it's the name of the envvar, I'd rename to something that ends in `_batch_size`, a `_count` is not very explicit

hook-common/src/pgqueue.rs
```diff
@@ -284,30 +316,93 @@ impl<J: std::marker::Send, M: std::marker::Send> PgQueueJob for PgJob<J, M> {
...
+        let mut txn_guard = self.shared_txn.lock().await;
 
-        self.transaction
-            .commit()
-            .await
-            .map_err(|error| PgJobError::TransactionError {
-                command: "COMMIT".to_owned(),
+        let txn_ref = txn_guard
+            .get_txn_ref()
+            .ok_or(PgJobError::TransactionAlreadyClosedError)?;
```
xvello, line 363: nit: you could avoid this by having `.get_txn_ref` return a Result instead of an Option. As the None case always leads to failure, that's fine?

hook-common/src/pgqueue.rs
```diff
@@ -284,30 +316,93 @@ impl<J: std::marker::Send, M: std::marker::Send> PgQueueJob for PgJob<J, M> {
...
+    }
+
+    /// See `raw_txn` above, this should be called when a transaction error occurs (e.g. the
+    /// connection drops) so that other jobs don't each try to do DB work that is bound to fail.
+    fn drop_txn(&mut self) {
+        self.raw_txn = None;
+    }
+}
+
+impl SharedTxn<'_> {
```
xvello: nit: while this is good engineering, we are in the "if I poke at this struct it might break in ways I don't yet understand" red zone, so I'd be fine dropping this.

sqlx transactions have [an `open` field set to false on `rollback()`](https://docs.rs/sqlx-core/0.7.3/src/sqlx_core/transaction.rs.html#92) so I'm assuming that the queries will fail-fast in that case. As `job.complete()` is simply an `update` query, I don't think there's a lot of value in handling that failure mode ourselves just yet.
I'd just have jobs call `tx.rollback()` on unrecoverable error, and let the rest propagate.

hook-worker/src/worker.rs
```diff
@@ -165,65 +172,87 @@ impl<'p> WebhookWorker<'p> {
...
-                let webhook_job = self.wait_for_job_tx().await;
-                spawn_webhook_job_processing_task(
-                    self.client.clone(),
-                    semaphore.clone(),
-                    self.retry_policy.clone(),
-                    webhook_job,
-                )
-                .await;
+                let mut batch = self.wait_for_jobs_tx().await;
+
```
xvello, line 191: nit, I'd love to have a histogram of the real batch size, to check whether we're hitting it and could benefit from raising it.

xvello (commented):
Looking good! Not approving as I think this deserves another pair of eyes.

I think we're digging our own grave with keeping two parallel implementations. I'd support dropping non-tx for now and switching to it later if we find that tx breaks some of our scalability assumptions.

Right now, I would not be comfortable switching prod to non-tx, so it's almost as good as dead code.

hook-common/src/pgqueue.rs
bretthoerner: Good call. Honestly, `SharedTxn` kind of fell out of my previous attempts to do the `commit` via `Drop`, and it does seem confusing and vestigial now with fresh eyes.

bretthoerner:
> I think we're digging our own grave with keeping two parallel implementations. I'd support dropping non-tx for now and switching to it later if we find that tx breaks some of our scalability assumptions.
> 
> Right now, I would not be comfortable switching prod to non-tx, so it's almost as good as dead code.

I would also prefer to just pick an implementation.

That said, I did have one more thought for us to consider. `PgQueue` is written as a generic queue. If we think we might use the queue for longer running jobs, I can see where holding PG transactions open for, say, 5 minutes (or longer, I'm just pulling numbers out of thin air) would get pretty sketchy. But transaction mode is pretty sweet for Webhooks where our longest timeout will probably be like 30 seconds (and most jobs will complete in hundreds of millis).

hook-common/src/pgqueue.rs
bretthoerner: [Removed it](https://github.com/PostHog/rusty-hook/pull/60/commits/8cb66e0a518fd695d162ce5fb3d06f2b55458daa), I think it makes things a lot easier to follow.

hook-worker/src/worker.rs
```diff
@@ -162,68 +177,94 @@ impl<'p> WebhookWorker<'p> {
...
-                    self.client.clone(),
-                    semaphore.clone(),
-                    self.retry_policy.clone(),
-                    webhook_job,
-                )
-                .await;
+                let mut batch = self.wait_for_jobs_tx().await;
+                dequeue_batch_size_histogram.record(batch.jobs.len() as f64);
+
+                // Get enough permits for the jobs before spawning a task.
```
xvello, line 192: We might be blocked here due to one missing permit.

We might get more throughput if we only pull `min(semaphore.available_permits(), dequeue_batch_size)` jobs, so that we guarantee we won't block here.

bretthoerner: Don't we want to block?

If I understand correctly and we use `min(semaphore.available_permits(), dequeue_batch_size)` then we could spawn off the work of 100 jobs by only grabbing 1 permit. If we do that then what's the point of having `max_concurrent_jobs` at all?

That said, I proposed `max_concurrent_jobs` because non-transaction mode has no coupling between PG connections and the amount of work in progress. We could drop this from the transaction mode branch since the connection limit and batch size gives you a max of work in progress. (Or hopefully we pick one mode and call it a day.)

xvello: Placed the comment here because this is where we block, but my recommendation applies higher: change the signature of `wait_for_jobs()` to accept a max job count to pull, so that we can call `wait_for_jobs(min(semaphore.available_permits(), dequeue_batch_size))`.

That's something we can implement later on, we can skip it for now and revisit if we manage to actually hit this semaphore.

bretthoerner: Ah-ha, gotcha, that makes sense.

xvello: Again, I think it's fine to just add a `TODO` comment when calling `wait_for_jobs()`, and move on for now.

bretthoerner:
@tomasfarias Any thoughts?

hook-common/src/pgqueue.rs
```diff
@@ -284,30 +315,56 @@ impl<J: std::marker::Send, M: std::marker::Send> PgQueueJob for PgJob<J, M> {
...
+    /// The open transaction the jobs in the Vec came from. This should be used to commit or
+    /// rollback when all of the work is finished.
+    shared_txn: Arc<Mutex<Option<sqlx::Transaction<'c, sqlx::postgres::Postgres>>>>,
+}
+
+impl<'c, J, M> PgTransactionBatch<'_, J, M> {
+    pub async fn commit(&mut self) -> PgQueueResult<()> {
+        let mut txn_guard = self.shared_txn.lock().await;
+
+        let txn = txn_guard.take().unwrap();
```
tomasfarias: What is this `take` call doing? From the `Mutex` [docs](https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#examples), it seems like we should deref `txn_guard`.

bretthoerner:
PR to remove non-transaction mode. It's a lot of noise so I wasn't sure if it would help or hurt reviewing this PR. Up to y'all: https://github.com/PostHog/rusty-hook/pull/65

hook-common/src/pgqueue.rs
tomasfarias: I'm familiar with the similarly named `std::mem::take`, and I'm afraid if we are replacing the txn held by the mutex.

bretthoerner: Commit takes ownership, so taking it from an `Option` is the only way I found to call it. I wouldn't say we're _replacing_ it but yes we are _taking_ it. Let me know if I'm missing something. :)

hook-worker/src/worker.rs
```diff
@@ -163,70 +178,104 @@ impl<'p> WebhookWorker<'p> {
...
+
+                let client = self.client.clone();
+                let retry_policy = self.retry_policy.clone();
+
+                tokio::spawn(async move {
+                    let mut futures = Vec::new();
+
+                    // We have to `take` the Vec of jobs from the batch to avoid a borrow checker
+                    // error below when we commit.
+                    for job in std::mem::take(&mut batch.jobs) {
```
tomasfarias, line 207: This makes me a bit nervous: The borrow checker is there to help us, not sure I want to _avoid_ it.

hook-worker/src/config.rs
```diff
@@ -37,6 +37,9 @@ pub struct Config {
 
     #[envconfig(default = "true")]
     pub transactional: bool,
+
+    #[envconfig(default = "10")]
```
tomasfarias: Any reason for the default? Otherwise I'd keep it in 1 until we need this.

hook-worker/src/worker.rs
bretthoerner: The borrow checker is still in play, maybe my comment is just bad. I can't find a way to take ownership of each job in the Vec (via `complete()` etc) *and* still be allowed to commit at the end.

This is just moving the `Vec` out of the batch, so we are the sole owner and can complete/fail jobs, and then the `batch` can still be used to commit.

hook-worker/src/config.rs
bretthoerner: Changed to 1 because I agree this is arbitrary, but I think we'll configure this to be non-0 in prod almost immediately. We currently only have 1 task dequeueing. If round trip to grab a job takes 5ms that's a 200 job/sec peak as written.

bretthoerner:
The current conflict will be un-conflicted when I re-merge the change I had to revert: https://github.com/PostHog/rusty-hook/pull/67

hook-common/src/pgqueue.rs
tomasfarias: Doesn't look like you are taking ownership, at least I see just a mutable reference to `PgTransactionBatch`. If you mean we take ownership to of the tx, we do, but in a way that requires leaving something behind in the `Option`. Why not take ownership of `PgTransactionBatch` instead? At least it seems it should be fine to be done with the batch once we commit.

I tried this out locally to see if there would be any compilation issues, and there are none (besides needing to bring the `sqlx::postgres::any::AnyConnectionBackend` trait into scope):

```rust
    pub async fn commit(self) -> PgQueueResult<()> {
        let mut txn_guard = self.shared_txn.lock().await;

        txn_guard
            .as_deref_mut()
            .ok_or(PgQueueError::TransactionAlreadyClosedError)?
            .commit()
            .await
            .map_err(|e| PgQueueError::QueryError {
                command: "COMMIT".to_owned(),
                error: e,
            })?;

        Ok(())
    }
```

tomasfarias: I kind of prefer this as sqlx takes ownership of transaction on commit, so it kind of makes sense that we take ownership of the transaction _batch_ when we commit it.

tomasfarias: Not very important though, I understood your explanation, thanks!

tomasfarias (approved):
Have a pending comment regarding ownership of `commit`, but it's not that critical.

Overall I'm fine with this. Still hesitant that we are prematurely optimizing, but after reading the PR a few times, I'm comfortable with moving forward. Ultimately we are saving pg connections, which is not a trivial thing to save, and I can see a future where we will end up batching at some point.

Thank you!

tomasfarias:
By the way, this PR will likely conflict with the error changes I'm doing at #36. Happy to get this merged out first @bretthoerner, and then I can rebase my PR and change the error types here too.

hook-common/src/pgqueue.rs
bretthoerner: > I kind of prefer this as sqlx takes ownership of transaction on commit, so it kind of makes sense that we take ownership of the transaction _batch_ when we commit it.

Yeah, makes sense to me, thanks!

PR #63, opened 2024-02-02
configurable max_connections for api/worker, default to 100

bretthoerner:
Janitor can keep the default (it only uses 1 at a time).


xvello (approved):

bretthoerner:
Merging this into https://github.com/PostHog/rusty-hook/pull/62 (but not into main)

PR #64, opened 2024-02-05
Add proper e2e histrogram based on metadata created_at

bretthoerner:


Cargo.lock
```diff
@@ -768,6 +768,7 @@ name = "hook-api"
 version = "0.1.0"
 dependencies = [
  "axum",
+ "chrono",
```
xvello, line 771: capture uses `time` instead as Ellie had concerns about `chrono` maintenance status, could we align on using `time` here too?

bretthoerner: That's interesting, news to me: https://github.com/chronotope/chrono/commits/main/

That said, while I had to pull it into this crate for the PR, `chrono` is already used elsewhere in the codebase. If you'd like us to switch then I'd rather do it as another PR since it will take a minute and seems unrelated.

hook-common/src/webhook.rs
```diff
@@ -135,6 +137,11 @@ pub struct WebhookJobMetadata {
     pub team_id: u32,
     pub plugin_id: i32,
     pub plugin_config_id: i32,
+    #[serde(
+        serialize_with = "serialize_datetime",
+        deserialize_with = "deserialize_datetime"
+    )]
+    pub created_at: DateTime<Utc>,
```
xvello, line 144: Are all fields of `WebhookJobMetadata` required? I'd argue that this one should be optional (and filled by api if not present)?

xvello (approved):

hook-common/src/webhook.rs
bretthoerner: They're required as currently written, yeah.

xvello: Ok for now, but I'm not sure we should ask all sources to fill this. I'd be fine with making it an `Option<DateTime<Utc>>` later on.

bretthoerner: Hm, the plugin-ness of the current code is expected, since the API contract very much matters to the janitor (for `app_metrics`). I guess the reason `Option` didn't come to mind here is that we only have 1 call-site that I've already updated.

Future non-plugin Webhook senders will require a little thought and probably their own endpoint, FWIW.

xvello: Yep, let's keep it that way, but I wanted to callout that we're making a breaking API change here.
Still fine for plugin-server only, we'll probably make another more generic API when opening to other use cases

PR #65, opened 2024-02-05
Remove non-transactional mode

bretthoerner:
This drops the non-transactional mode, i.e. any usage of the `running` status.

Note that I didn't add a migration to drop the `running` status yet, changing an enum that's in use is moderately complicated in PG. It'd be better to do the next time we make a new shard (if we even care).

What I think we'd have to do (not worth it):
```
CREATE TYPE job_status_new AS ENUM('available', 'completed', 'failed');
ALTER TABLE job_queue ADD COLUMN status_new job_status_new;
-- Here we'd need to set all rows to the proper status, and make sure we get it right with transaction in flight and locked rows, etc?
ALTER TABLE job_queue DROP COLUMN status;
ALTER TABLE job_queue RENAME COLUMN status_new TO status;
DROP TYPE job_status;
ALTER TYPE job_status_new RENAME TO job_status;
```

### DIFF
--- a/hook-common/src/pgqueue.rs
+++ b/hook-common/src/pgqueue.rs
@@ -60,8 +60,6 @@ pub enum JobStatus {
     Discarded,
     /// A job that was unsuccessfully completed by a worker.
     Failed,
-    /// A job that was picked up by a worker and it's currentlly being run.
-    Running,
 }
 
 /// Allow casting JobStatus from strings.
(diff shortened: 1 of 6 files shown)


xvello:
Re: schema: I'd support keeping the column for now (and marking it as vestigial). If we find out that non-tx is needed, we would not have to change the schema back. As it's an enum, the overhead should be pretty small.

tomasfarias (approved):
Happy to keep the enum for the time being, don't think it's much of an overhead :+1:

PR #67, opened 2024-02-05
Add histogram for insertion time to completion time

bretthoerner:
~~Doh, `plugin-server` is deployed and sends the `created_at` field, but I forgot to consider that `hooks-api` wouldn't be serializing it into the DB until deployed, and so the race of deploying `hooks-api` and `hooks-worker` left rows that couldn't be deserialized in the DB.~~

~~Making this an `Option` as this field isn't critical to normal operation. I go ahead and set it in `hooks-api` if it's missing.~~

~~@xvello was right :)~~

Based on `created_at` column now.

### DIFF
--- a/hook-worker/src/worker.rs
+++ b/hook-worker/src/worker.rs
@@ -2,6 +2,7 @@ use std::collections;
 use std::sync::Arc;
 use std::time;
 
+use chrono::Utc;
 use hook_common::health::HealthHandle;
 use hook_common::{
     pgqueue::{Job, PgJob, PgJobError, PgQueue, PgQueueError, PgQueueJob, PgTransactionJob},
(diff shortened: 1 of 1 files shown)


xvello:
The Job struct already has the [created_at](https://github.com/PostHog/rusty-hook/blob/0426a8196d89c80d3831c1b98d6864b4211a0062/hook-common/src/pgqueue.rs#L91) field sourced from PG, I suggest using it directly. As the submission through hooks-api is synchronous, its value should be pretty close to the source-submitted metadata. I am also wondering about the confusion factor of having a `created_at` column in the table + a `created_at` metadata, and what they mean.

bretthoerner:
> As the submission through hooks-api is synchronous

FWIW, my thinking was that a real end-to-end test takes into account any difficulties talking to the API (or if the API is blocked waiting on a PG connection from its pool, etc). In the end though I'm indifferent, and agree two `created_at` was a bad choice.

bretthoerner:
Moving this to new repo to try out actions etc. https://github.com/PostHog/hog-rs/pull/5

PR #68, opened 2024-02-06
fix: Panic on exceeding max retries

tomasfarias:
As it stands, the only thing stopping us from retrying jobs indefinitely is an `if` check in the `retry` method. However, future refactors may omit this check for one reason or another. 

Ideally, the transition to a `RetryableJob` should fail and give the caller the chance to recover by failing the job instead. The reason we don't do this is that we return the wrapper `PgJob`/`PgTransactionJob` in the event a `retry` is not possible so that it may be failed by the worker instead. However, we require ownership of the underlying `job` when transitioning to a `RetryableJob`, so if we were to return the wrapper `PgJob` or `PgTransactionJob` when the transition fails it would be too late as partial ownership has already been moved to `retryable`. So, we do not error on the transition but instead enforce it with an if check.

To make things safer, I've added a panic in the transition. This way, we'll crash upon forgetting to check if the job can be retried. While we experiment on how to change this, at least the panic will prevent potential data corruption (for example, a job is updated with attempts beyond its max attempts).

Naturally, I'm not super thrilled about this approach: Although safer, I would like to give callers the chance to recover (by failing the job instead of recovering) which we are not doing with a panic. 

An idea I'm consedering is including retrying as part of `fail` (perhaps renaming the method to something like `retry_or_fail`). This way, we don't have to return the `PgJob`/`PgTransactionJob` as the recovery from a failed retry (failing the job) will be executed immediately after without requiring the caller to do it and thus requiring ownership to be moved back to the caller. The caller can check if the return type is a `RetriedJob` or a `FailedJob` to deduce what transition happened.

Any thoughts?

### DIFF
--- a/hook-api/src/handlers/app.rs
+++ b/hook-api/src/handlers/app.rs
@@ -30,9 +30,7 @@ mod tests {
 
     #[sqlx::test(migrations = "../migrations")]
     async fn index(db: PgPool) {
-        let pg_queue = PgQueue::new_from_pool("test_index", db)
-            .await
-            .expect("failed to construct pg_queue");
+        let pg_queue = PgQueue::new_from_pool("test_index", db).await;
 
         let app = add_routes(Router::new(), pg_queue);
 
--- a/hook-api/src/handlers/webhook.rs
+++ b/hook-api/src/handlers/webhook.rs
@@ -127,9 +127,7 @@ mod tests {
 
     #[sqlx::test(migrations = "../migrations")]
     async fn webhook_success(db: PgPool) {
-        let pg_queue = PgQueue::new_from_pool("test_index", db)
-            .await
-            .expect("failed to construct pg_queue");
+        let pg_queue = PgQueue::new_from_pool("test_index", db).await;
 
         let app = add_routes(Router::new(), pg_queue);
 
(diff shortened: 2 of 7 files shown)


bretthoerner:
`retry_or_fail` sounds reasonable to me, maybe I'm not thinking of a downside?

I always forget the details, but I guess `panic` in async Rust just kills the task. I'm trying to think of what would happen in practice in this situation. I guess we'd keep `dequeuing` the job, most likely failing, and then panicing again if we had a retry count bug?

If I'm understanding right, it seems like we'd want a metric and an alert because I doubt we'd notice this in the logs? But if  `retry_or_fail` is a proper fix then that sounds better.

tomasfarias:
> I always forget the details, but I guess panic in async Rust just kills the task. I'm trying to think of what would happen in practice in this situation. I guess we'd keep dequeuing the job, most likely failing, and then panicing again if we had a retry count bug?

Yeah, in the event there is a bug that causes us to attempt to retry a job that cannot be retried further we would panic and just kill the task. Job would still be in the db though, like you said, so we would keep dequeuing and  keep hitting the bug (unless the bug is unstable for some reason).

> If I'm understanding right, it seems like we'd want a metric and an alert because I doubt we'd notice this in the logs? But if retry_or_fail is a proper fix then that sounds better.

Yeah, me panicking here is more of a safety patch, an api change (i.e. `retry_or_fail`) is I think the way to work around this. I can add that instead of the panic. Been struggling to come up with a solution that maintains the current API, but I just haven't been able to work around ownership (maybe for a good reason).

bretthoerner:
Closing this to archive. It sounds like you were making a different change anyway, which we can put at https://github.com/PostHog/hog-rs

